PR #37, opened 2016-01-21
Update Minion_Job_DigMineStairwell.java

BorisSFR:
The added code will add cobblestone walls to the stairwell/staircase.


BorisSFR:
Where ever the grid is generated needs to be updated to a 7x7 so it will show that the walls are included...

AtomicStryker:
The relevant line would be atomicstryker/minions/client/MinionsClient.java:167

BorisSFR:
Thanks Atomic I have updated that file with suggested changes as well.

BorisSFR:
Atomic, in the MinionsClient.java I have also added commented code that would offset the grid a different way.  If you like this suggestion better I can reedit my suggestion in the DigMineStairwell to accomplish this.

PR #48, opened 2016-03-22
1.7.10

homacs:
Hello AtomicStryker! 

Finished work on multi-threading. Could have done more but wanted to keep changes simple. Modified basically FileHandler and RuinGenerator to get them thread-safe. Also added a config entry which allows to choose between multi-threaded and single-threaded. Multi-threaded is the default case. 

Have a look if it fits in.


AtomicStryker:
Ok this pull request is unfortunately unusable, like this.
Please don't feel discouraged, i love your effort, but reasons:

1) [main reason, formalities]
You break formatting, the actual changes are almost invisible in a tide of spaces-to-tabs or something changes. Also i spotted atleast one case of mixing brace styles (brace on same line instead of what i use). It's not that i am a zealot about one style over the other, but there should be one consistent style and i even have a formatter xml in my workspace.
Also, there is atleast one unused import after your changes.

2) [more formalities]
You refactored/renamed variables and methods out of personal preference? That's not nice. You also changed the access levels on a bunch (yes i know, IDEA suggested this). If your renamed choices were clearly superior or something but "isDisableLogging()", really?

3)
To what you actually changed, the way i see it: 
a) making the whole async loader optionally sync [valid change i guess?]
b) bugfix moving the map clear into an earlier event [good change]
c) adding synchronized keyword to flush method, also adding a modified flag to skip unnecessary writing [good change, there is actually a bug here with simultaneous writers]
d) adding synchronized keyword in a couple of places in the templace file loading process, adding wait() to bounce the thread lock to the loaders, making the concurrent list an arraylist again. Also adding error propagation. My question to you: Why so many synchronized locks? Each adds overhead, wouldn't it be better to just have one loader thread running max, with the others waiting their turn? This would only require a single synchronized or lock per dimension, the way i see it.

4)
The 1.7.10 branch is basically legacy at this point and no longer actively developed. I only ported bugfixes because people still use it and im a sucker.

Again, please don't feel like im bashing you or your effort. I am not. I haven't managed to get a forge PR past Lex either, it's some kind of ritual.

PR #51, opened 2016-03-26
Fix crash when running deobf Forge

plnqux:
This PR fixes a crash that occurs when trying to run a deobf Forge with DynamicLights imported as a dependency.


AtomicStryker:
Excellent.

PR #83, opened 2017-01-24
Fix for ender mod not breaking out of loop if it gets under y = 0

Corosauce:
Was spam fighting endermen from a spawner deep in a roguelike dungeon, around y 15. Encountered enderman with Ender modifier, multiple times upon hit, server would freeze, then time out. I'm guessing the starting teleport point it randomized to had a y of less or near 0 and then kept going, I believe this PR should fix that. Issue probably exists as far back as mc 1.7.10 as that is the version I encountered it on.


AtomicStryker:
Thank you, obvious improvement is obvious

PR #103, opened 2017-04-28
Ore dictionary for custom needles

zetti68:
Hello AtomicStryker,
Sorry I was to stupid. I didn't fork the project, I cloned it. Thanks for your help, I found the right function.

New/ changed function:
1. Search block for given name
1.1 If not found, search in ore dictionary with given name
2. Search ore dictionary name of block (only if block found)
3. Add all block to CompassTargetData -> contains a list of blocks now
4. Search the nearest block of group and use it as needle target

One change was necessary: use damage 0 as placeholder and not -1. -1 is not a legal damage in ore dictionary.
Todo:
Combine wildcard -1 and 0.


FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
```diff
@@ -204,4 +217,56 @@ else if (settingList.isEmpty())
...
+				if (!compassTargetData.isEmpty()){
+					data.put(oreDictName, compassTargetData);
+				}
+			}
+		} else {
+			String oreDictName = pBlockName;
+			CompassTargetData compassTargetData = new CompassTargetData(oreDictName);			
+			for (ItemStack stack : OreDictionary.getOres(oreDictName)) {
+				Item item = stack.getItem();
+				int damage = item.getDamage(stack);
```
AtomicStryker: You put all meta values in? I don't know enough about the Oredict, but the function does take meta arguments - can there be different (meta) blocks behind an oredict name?

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
```diff
@@ -162,22 +170,27 @@ else if (settingList.isEmpty())
                             }
                             else
                             {
-                                configInts[8] = -1;
+                                configInts[8] = 0;
                             }
```
AtomicStryker: I am fairly sure the -1 meant something (over 0), did you adjust for that?

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
AtomicStryker: Ah, i read the changes in the config doc. Question answered.

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
```diff
@@ -78,7 +85,7 @@ public void writeDefaultFile(File fileToPrint)
             pw.println("NoEnderEyeNeedle");
             pw.println("//");
             pw.println("// this is Diamond = block id 56, with a needle color of {51,255,204}, it scans 15 blocks horizontally, 1 vertically, from 1-16 height, every second");
-            pw.println("diamond_ore:51:255:204:15:1:1:16:0");
+            pw.println("oreDiamond:51:255:204:15:1:1:16:0");
             pw.println("//");
```
AtomicStryker, line 92: Just so i don't get it wrong - the old names still work without changes, right? I do not want the QQ of users whose configs no longer work

FinderCompass/src/main/java/atomicstryker/findercompass/common/CompassTargetData.java
```diff
@@ -29,14 +64,53 @@ public boolean equals(Object o)
...
+            	BlockData comp = (BlockData)o;
+                return comp.getBlockID() == blockID && comp.getDamage() == damage;
+            }
+            return false;
+        }
+        
+        @Override
+        public int hashCode()
+        {
+            return blockID.getUnlocalizedName().hashCode();
```
AtomicStryker: I am not 100% but i think i've had this return null on mod blocks, before

FinderCompass/src/main/java/atomicstryker/findercompass/client/ThreadCompassWorker.java
```diff
@@ -1 +1 @@
...
    }
    
    private double getDistanceBetweenChunks(ChunkCoordinates var1, ChunkCoordinates var2)
    {
        int xdiff = Math.abs(var1.posX - var2.posX);
        int ydiff = Math.abs(var1.posY - var2.posY);
        int zdiff = Math.abs(var1.posZ - var2.posZ);
        return Math.sqrt(Math.pow((double)xdiff, 2.0D) + Math.pow((double)ydiff, 2.0D) + Math.pow((double)zdiff, 2.0D));
    }
}
```
AtomicStryker: Well, something went wrong here. See if you can commit this in readable format?

FinderCompass/src/main/java/atomicstryker/findercompass/client/CompassSetting.java
```diff
@@ -56,6 +57,21 @@ public boolean isStrongholdNeedleEnabled()
         return customNeedles;
     }
     
+    public CompassTargetData getCustomNeedle(String pOreDictName)
+    {
+    	CompassTargetData needle = null;
+    	if (pOreDictName != null){
+    		for (CompassTargetData entry : customNeedles.keySet()) {
+				if (entry.getOreDictName().equals(pOreDictName))
```
AtomicStryker: Is this never null? Otherwise flipping entry.getOreDictName() and pOreDictName prevents theoretical NPE

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
AtomicStryker: I've now read the description stating this. Uh so what about users who have -1 in their configs?

AtomicStryker (changes requested):
Please fix the class that is displayed incorrectly (line endings probably), otherwise some nice work, if the config is downward compatible i will merge this probably.

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
zetti68: Old and new names are working. It is only an example to show the new possible name.

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
zetti68: I have an idea to combine the special handling and the ore dictionary functionallity. -1 isn't the real problem, it is easy to change it to 0 automatically. But the problem is the free definition of damage.
My idea is to use a boolean flag (ignoreDamage) in CompassTargetData if no or -1 damage is set. In this case it is necessary to read the dictionary block names from damage 0 to upper with step width 1. If the ore dictionary name changes or there isn't a block, it will stop.
I think it isn't a good idea to search all blocks with different damage and different ore dictionary names for one identical needle.

FinderCompass/src/main/java/atomicstryker/findercompass/common/CompassTargetData.java
zetti68: I'm not sure, if it is possible or not to get a null return. I'll check the source in forge and optional add a second function.

FinderCompass/src/main/java/atomicstryker/findercompass/client/ThreadCompassWorker.java
zetti68: Your source file of this class is formatted as unix file. SourceTree cannot check the differences correctly.

FinderCompass/src/main/java/atomicstryker/findercompass/client/CompassSetting.java
zetti68: Right, done.

FinderCompass/src/main/java/atomicstryker/findercompass/common/CompassTargetData.java
zetti68: NPE isn't possible. Method returns "tile." + mUnlocalizedBlockName.

AtomicStryker: Uuunless a mod block overrides the method

FinderCompass/src/main/java/atomicstryker/findercompass/client/ThreadCompassWorker.java
AtomicStryker: My email client displays this correctly, for whatever reason. So i've checked out your critical "inner loop" and it is a bit inefficient - instead of storing the world.getBlock result and comparing it against the blocklist, you call getBlock once for every blocklist entry. Let's assume you have something like mod copper with 5 types, that is 4 superfluous getBlock calls per block, in a 16 by 16 area, 4 blocks height, that's a measly 4096 getBlock calls that could have been avoided. Please amend this.

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
AtomicStryker: I am still confused. Is it compatible with people who have -1 set or not? Explain it like i'm really really dense

FinderCompass/src/main/java/atomicstryker/findercompass/client/ThreadCompassWorker.java
```diff
@@ -1 +1,123 @@
...
+        {
+            if (yIter >= minY && yIter <= maxY)
+            {
+                for (int zIter = playerZ - xzRange; zIter <= playerZ + xzRange; ++zIter)
+                {
+                    for (int xIter = playerX - xzRange; xIter <= playerX + xzRange; ++xIter)
+                    {
+                    	for (BlockData blockData : blocks)
+                    	{
+                    		if (this.mcinstance.theWorld.getBlock(xIter, yIter, zIter) == blockData.getBlockID())
```
AtomicStryker: This spot is what i mean by "inner loop". You are re-getting the getBlock result when you should be saving it in a local variable BEFORE iterating "blocks"

FinderCompass/src/main/java/atomicstryker/findercompass/common/CompassTargetData.java
zetti68: Changed completely, NPE isn't possible

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
zetti68: Last version is fully compatible, but -1 isn't a legal damage in minecraft. If the old configuration file contains a damage -1, it will be set to 0. With -1 it is not possible to find a block in ore dictionary.
The new searching is:
- Search all IDs in ore dictionary for given block. If a damage is set it will be used. Otherwise it searches all IDs for different damages of same block.
- Create CompassTargetData for all located blocks.
- It also works for block, these are not registered in ore dictionary.

New function during searching:
- Use flag useDamage instead check damage against -1 in ThreadCompassWorker

AtomicStryker: I know it isn't, i had to add it because someone complained about using 0 as wildcard, except he wanted to find only damage 0 blocks

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
```diff
@@ -290,8 +292,14 @@ else if (settingList.isEmpty())
 			} else {
 				int[] ids;
 				for(int damage = 0; (ids=OreDictionary.getOreIDs(new ItemStack(pBlock, 1, damage))).length > 0; damage++){
+					boolean added = false;
 					for(int id:ids){
+						int size = oreDictIDs.size();
 						oreDictIDs.add(id);
+						added = added || size != oreDictIDs.size();
```
AtomicStryker: Wait what. You just added something. Why would size NOT change?

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
```diff
@@ -290,8 +292,14 @@ else if (settingList.isEmpty())
 			} else {
 				int[] ids;
 				for(int damage = 0; (ids=OreDictionary.getOreIDs(new ItemStack(pBlock, 1, damage))).length > 0; damage++){
```
AtomicStryker: You are blindly iterating damage values upward ... what if there are gaps? What if all items of any damage match the oredict entry? I would search for the infinite loop here

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
zetti68: This was the fastest way to stop the endless loop. My first code works with a few mods too, but we have a server with 250 mods and where was the endless loop.
It is not easy to debug this. The only reason for the endless loop is the return of the same set of IDs of same block with different damage. And this will get ever and ever for each damage (from 0 to endless).
If all IDs are already in the set, nothing will be added and the size is the same (HashSet) before and after add. More logical is the check with contains(), but size check should be (a little bit) faster.

FinderCompass/src/main/java/atomicstryker/findercompass/common/DefaultConfigFilePrinter.java
zetti68: A gap will break the loop and if there are different damages for one ore dictionary block, it will return only one id.

Please see comment above. This is not necessary, if we work without the placeholder -1. Normally has a block only one damage and this default is 0. If you want to use a placeholder to find all blocks of same block with different damages, it is necessary to iterate over all defined damages.
It would be logical if block with a special damage isn't registered in ore dictionary, the given set of IDs is empty and so the loop will stop. It works for a few mods without problems. But there are mods these give for any damage of a block the same set of IDs. And so I need the additional check.
It is also possible to delete the new function completely. But there is the function to search with the wildcard -1 a big set of different damages of same block at this time. And this isn't possible without the iteration over damage in ore dictionary.

zetti68: I have another idea. It is also possible to ignore ore dictionary if -1 is explicitly set as damage. In this case it will only uses the block name. It searchs in ore dictionary without or with damage >= 0.

Please let me know, if it could be fine. In this case I will change the source.

AtomicStryker: Sorry, i am currently swamped by pesky real life and not giving this the attention it deserves. Yes, this sounds like a good idea. I don't expect any sizable amount of users to actually use -1, but still, breaking backwards compatibility is bad.

zetti68:
Hello AtomicStryker,
I hope I changed the search algorithm to a more logical version. It combines the old wildcard function (->ignore damage) with the searching in ore dictionary.

There are four strategies to define a needle:
1. Define block with block name without damage:
-> add given block with all damages
-> search block with damage 0 in ore dictionary and add all registered blocks to one needle
2. Define block with block name with damage
-> add given block with identical damage
-> search block with given damage in ore dictionary and add all registered block to one needle
3. Define block with block name with damage <= -1
-> add only block with all damages to one needle
-> do not search in ore dictionary
4. Define ore dictionary name
-> search all blocks in ore dictionary for given name and add these to one needle
-> ignores defined damage

AtomicStryker (approved):

AtomicStryker:
I'm leaving this unmerged for a couple more days in the hopes (and nothing more) that you and your modpack buddies playtest a bit more - i don't have the time. I don't think i will have the time to immediatly port the changes over to 1.11.2 either. Great job anyway, this is like the third pull request i've had in half a decade and by far the biggest.

zetti68:
Thanks for integration. If I find a little bit time, I will port it to newer versions of minecraft.
I found a little bug, it is necessary to check mUseDamage in method equals too. Sorry for this bug fix request.

PR #113, opened 2017-06-02
1.11.2 Exclude Nether & End blocks for "generic" templates.

MentalMouse:
Hand-edited, because I'm not so eager for a local clone of AS's entire world.


AtomicStryker:
Why thanks, those look useful.

PR #121, opened 2017-06-28
Adding more control for spawning randomized structures.

staalek:
Added option to spawn an adjoining template with a relative Y.
Added option to randomly pick one of a set of adjoining templates.

New syntax:
adjoining_template=&lt;template&gt;[,&lt;template&gt;[,..]];&lt;relativeX&gt;;&lt;allowedYdifference&gt;;&lt;relativeZ&gt;[;&lt;spawnchance&gt;[;&lt;useRelativeY&gt;;&lt;relativeY&gt;]]
Same as with block rules, specify same template multiple times to change weight.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
@@ -7,6 +7,7 @@ import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.Random;
 
@@ -39,14 +40,18 @@ public class RuinTemplate
     private boolean preventRotation = false;
     private final ArrayList<Integer> bonemealMarkers;
     private final ArrayList<AdjoiningTemplateData> adjoiningTemplates;
+    
+    private static Hashtable<String,RuinTemplate> knownTemplates = new Hashtable<>();
 
     private class AdjoiningTemplateData
     {
-        RuinTemplate adjoiningTemplate;
+        ArrayList<RuinTemplate> adjoiningTemplates;
         int relativeX;
         int acceptableY;
         int relativeZ;
         float spawnchance;
+        boolean useRelativeY;
+        int relativeY;
     }
 
     public RuinTemplate(PrintWriter out, String filename, String simpleName, boolean debug) throws Exception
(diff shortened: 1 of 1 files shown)


Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
```diff
@@ -743,22 +752,41 @@ else if (line.startsWith("preventRotation"))
                 }
                 else if (line.startsWith("adjoining_template"))
                 {
-                    // syntax: adjoining_template=<template>;<relativeX>;<allowedYdifference>;<relativeZ>[;<spawnchance>]
+                    // syntax: adjoining_template=<template>[,<template>[,..]];<relativeX>;<allowedYdifference>;<relativeZ>[;<spawnchance>[;<useRelativeY>;<relativeY>]
```
AtomicStryker, line 755: Why the addition of "useRelativeY" at all? Simply use a default value of relativeY=0 to cover the useRelativeY=0 case?

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
```diff
@@ -465,26 +470,30 @@ else if (curRule.runLast())
...
             float randres = (world.rand.nextFloat() * 100);
             if (randres < ad.spawnchance)
             {
                 int newrot = world.rand.nextInt(4);
                 int targetX = xBase + ad.relativeX;
                 int targetZ = zBase + ad.relativeZ;
-                int targetY = ad.adjoiningTemplate.checkArea(world, targetX, y, targetZ, newrot, ad.acceptableY);
-                if (targetY > 0 && Math.abs(y - targetY) <= ad.acceptableY)
+                int targetY;
+                if (ad.useRelativeY) targetY = y + ad.relativeY;
```
AtomicStryker, line 483: Minor point, but you break the formatting rules here (i have a formatter in my workspace, no? Yes i do)

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
```diff
@@ -39,14 +40,18 @@
     private boolean preventRotation = false;
     private final ArrayList<Integer> bonemealMarkers;
     private final ArrayList<AdjoiningTemplateData> adjoiningTemplates;
+    
+    private static Hashtable<String,RuinTemplate> knownTemplates = new Hashtable<>();
```
AtomicStryker, line 44: Instead of this, why not refactor FileHandler.templates into something static to get access to instances? Have the Ruin file Loader run once over all templates without actually instantiating the adjoining templates, then "connect" them by reference. You are essentially keeping a second set of RuinTemplate objects here.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
```diff
@@ -743,22 +752,41 @@ else if (line.startsWith("preventRotation"))
...
+                    	if (knownTemplates.containsKey(templateName))
+                    	{
+                    		data.adjoiningTemplates.add(knownTemplates.get(templateName));
+                    	}
+                    	else
+                    	{
+		                    File file = new File(RuinsMod.getMinecraftBaseDir(), "mods/resources/ruins/" + templateName + ".tml");
+		                    if (file.exists() && file.canRead())
+		                    {
+		                        RuinTemplate adjTempl = new RuinTemplate(debugPrinter, file.getCanonicalPath(), file.getName(), false);
```
AtomicStryker, line 779: Something something infinite recursion

AtomicStryker (changes requested):
The idea itself is fine and i think it could be merged. The code is mostly fine, except for the possible infinite recursion through misconfiguration, but maybe you can do better? 🍡

staalek:
Thanks for feedback, I'll get right on it
Though, the useRelativeY is so that you can actually use the relative Y value of 0, in case you want to expand outward on the same height (and not have it autoshift up and down according to terrain)
It can do infinite recursion, but you already have a warning about that ;)
(# these will also not be checked for circular dependencies, you build an infinite loop, you wait it out)

PR #136, opened 2017-08-25
new template rule features--see template_rules.txt

QuarterAnimal:
I needed per-structure random draws in addition to per-block (so, for example, the walls of some structures might be all oak planks and others all spruce, and the half the oak structures might have cobblestone floors, while the other half and all the spruce ones have dirt...stuff like that). With the current rules, the best you can do is get a jumble of oak and spruce and cobble and dirt in every structure, or make separate templates for each desired combination. My solution was "rule variants"; maybe this has general usefulness.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
@@ -26,132 +26,6 @@ import net.minecraftforge.common.MinecraftForge;
 
 public class RuinTemplate
 {
-    private static class VariantRuleset
-    {
-        private static class VariantGroup
-        {
-            private static class VariantRule
-            {
-                private final ArrayList<Integer> weights;
-                private int weightsTotal;
-                private final ArrayList<RuinTemplateRule> variants;
-
-                public VariantRule(final int weight, final RuinTemplateRule variant)
-                {
-                    weights = new ArrayList<>();
-                    weightsTotal = 0;
-                    variants = new ArrayList<>();
-                    addVariant(weight, variant);
-                }
-
-                public void addVariant(final int weight, final RuinTemplateRule variant)
-                {
-                    weights.add(weight);
-                    weightsTotal += weight;
-                    variants.add(variant);
-                }
-
-                public int getRandomSelector(Random random)
-                {
-                    return random.nextInt(weightsTotal);
-                }
-
-                public RuinTemplateRule getVariant(final int selectorInitial)
-                {
-                    RuinTemplateRule variant = variants.get(variants.size() - 1);
-                    if (selectorInitial < weightsTotal)
-                    {
-                        int index = 0;
-                        for (int selector = selectorInitial; (selector -= weights.get(index)) >= 0; ++index);
-                        variant =  variants.get(index);
-                    }
-                    return variant;
-                }
-            }
-
-            private final int repeatCount;
-            private final ArrayList<VariantRule> variantRules;
-
-            VariantGroup(final int repeatCountSpec, final int weight, final RuinTemplateRule variant)
-            {
-                repeatCount = repeatCountSpec;
-                variantRules = new ArrayList<>();
-                addVariantRule(weight, variant);
-            }
-
-            public void addVariant(final int weight, final RuinTemplateRule variant)
-            {
-                variantRules.get(variantRules.size() - 1).addVariant(weight,  variant);
-            }
-
-            public void addVariantRule(final int weight, final RuinTemplateRule variant)
-            {
-                variantRules.add(new VariantRule(weight, variant));
-            }
-
-            public ArrayList<RuinTemplateRule> getVariants(Random random)
-            {
-                final ArrayList<RuinTemplateRule> variants = new ArrayList<>();
-                for (int i = 0; i < repeatCount; ++i)
-                {
-                    final int selector = variantRules.get(0).getRandomSelector(random);
-                    for (final VariantRule variantRule: variantRules)
-                    {
-                        variants.add(variantRule.getVariant(selector));
-                    }
-                }
-                return variants;
-            }
-
-            public int size()
-            {
-                return repeatCount*variantRules.size();
-            }
-        }
-
-        private final ArrayList<VariantGroup> variantGroups;
-
-        VariantRuleset()
-        {
-            variantGroups = new ArrayList<>();
-        }
-
-        public void addVariant(final int weight, final RuinTemplateRule variant)
-        {
-            variantGroups.get(variantGroups.size() - 1).addVariant(weight,  variant);
-        }
-
-        public void addVariantRule(final int weight, final RuinTemplateRule variant)
-        {
-            variantGroups.get(variantGroups.size() - 1).addVariantRule(weight,  variant);
-        }
-
-        public void addVariantGroup(final int repeatCount, final int weight, final RuinTemplateRule variant)
-        {
-            variantGroups.add(new VariantGroup(repeatCount, weight, variant));
-        }
-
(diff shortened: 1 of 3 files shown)


Ruins/src/main/resources/examplesAndDocs/template_rules.txt
```diff
@@ -284,6 +284,117 @@ preserve_plants=0
...
+# now be used. This can make rules shorter and, in certain circumstances, more efficient. Apply a weight to an ID in a rule by
+# preceding it with a prefix of the form "<weight>*"; this is functionally equivalent to repeating the ID <weight> times. Valid
+# values are integers between 1 and 99999, inclusive. For example, the following rules are essentially identical, each producing
+# blocks of cobble, mossy cobble, and gravel with roughly 56%, 33%, and 11% likelihood, respectively:
+#     rule1=0,100,cobblestone,cobblestone,mossy_cobblestone,cobblestone,gravel,cobblestone,mossy_cobblestone,mossy_cobblestone,cobblestone
+#     rule2=0,100,5*cobblestone,gravel,3*mossy_cobblestone
+# Weights are optional. If no weight is specified for an ID in a rule, a weight of 1 is assumed.
+#
+# Rule Variants: Consider a platform composed of many instances of the following rule3:
+#     rule3=0,100,wool-14,wool-4,wool-11
```
AtomicStryker: You should write down what effect each meta stands for so someone who doesn't know ruins can figure out the math
I mean, i barely grasp it and i know how the weights work

Ruins/src/main/resources/examplesAndDocs/template_rules.txt
```diff
@@ -284,6 +284,117 @@ preserve_plants=0
...
+#     4*rule9=5*0,100,stone
+#     rule10^4*0,100,8*stone,cobblestone
+#     ^0,100,preserveBlock
+#     rule11^3*0,100,4*stone,cobblestone
+#     ^0,100,preserveBlock
+#     rule12^2*0,100,2*stone,cobblestone
+#     ^0,100,preserveBlock
+#     rule13^0,100,stone,cobblestone
+#     ^0,100,preserveBlock
+# This creates rule9 through rule13 as before, but also creates rule14 through rule28, so you don't have to.
```
AtomicStryker: 2 characters difference creates a whole lot of "invisible" rules ... that's unfortunate

What happens when rule ID overlap?

Ruins/src/main/resources/examplesAndDocs/template_rules.txt
```diff
@@ -284,6 +284,117 @@ preserve_plants=0
...
+#     ^2*0,100,log-2
+#     rule8^5*0,100,planks-0
+#     ^0,100,planks-1
+#     ^2*0,100,planks-2
+# Some of your pillars (5/8) will be made of oak, fewer (1/4) of birch, and fewer still (1/8) spruce. However, because the weights
+# are identical in these two rules, the logs and planks are always of the same wood type. There's no law that says the weights
+# have to be identical, though; this can be exploited to interesting effect. Consider stacks of the following rules, once again
+# from the bottom up, one block of each:
+#     rule9=5*0,100,stone
+#     rule10^4*0,100,8*stone,cobblestone
```
AtomicStryker: What happens when i "place" named variant rules? In this example the pillar is all "9", but what if i use "10" for example

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
```diff
@@ -564,26 +692,26 @@ private void levelSite(World world, Block fillBlockID, int xBase, int y, int zBa
...
         parseVariables(lines);
 
-        // the first rule added will always be the preserve block rule.
-        rules.add(new RuinRuleAir(debugPrinter, this));
-
         // now get the rest of the data
         final Iterator<String> i = lines.iterator();
         String line;
-        int ruleCount = -1;
+        int ruleState = 0;
```
AtomicStryker: This needs to be an enum, this number is magic

AtomicStryker (commented):
First, congratulations. This PR makes me feel inadequate. It is the greatest thing i've seen requested of me, ever.

Major downside for Ruins being the variant rules are approaching mathematical complexity. Debugging those will be fuuuuun.

On the matter of debugging: There is something very important missing, IMO. You need to be extremely verbose about what is happening. When you throw exceptions, specify the rule and template. Use debugPrinter to write about every little bit of logic happening (creating invisible anonymous variant rules, creating named variant rules, GENERATING(!!!) variant rules...).

That way, if there's a template issue (and there will be), a user testing the template will hopefully figure out what is wrong with his mess.

Ruins/src/main/resources/examplesAndDocs/template_rules.txt
QuarterAnimal: In this example, the pillar isn't all 9 (well...it COULD be, but then it'd be all stone, since 9 is always stone). The intent was to make the base 9, the block above it 10, the one above that 11, and so on. Named rules always create numbers you can use in your level, unnamed ones never do. Here, 10 isn't a variant of 9--it's its own separate rule, but is "linked" to 9 with respect to which of its variants is selected for a particular instantiation of the template.

Ruins/src/main/resources/examplesAndDocs/template_rules.txt
QuarterAnimal: As always, the actual rule ID is assigned based entirely on its ordinal position in the template file--the first rule is 1, the second 2, and so on. The "names" are arbitrary and irrelevant; you can name all your rules "rule666" if you want to, and they'll still have IDs 1, 2, etc. Therefore, IDs can't possibly overlap, even if you try. I'm not particularly crazy about that, but I left it as is to maintain backward compatibility.

As for the repeat count...yeah, it's kind of hinky. I find it useful, but I wouldn't b averse to dropping that feature. It doesn't really add a new capability; you can accomplish exactly the same thing with cut-and-paste. The convenience may not be worth the potential for confusion.

Ruins/src/main/resources/examplesAndDocs/template_rules.txt
QuarterAnimal: I'm not sure I understand. In this rule, the 14, 4, and 11 are standard Minecraft wool color metadata values for red, yellow, and blue. They've got nothing to do with weights. This rule doesn't use any of the features I added; it (and the accompanying text) is only here to illustrate the difference between per-block random selection (what the current mod provides) and per-structure random selection (the whole motivation for rule variants).

AtomicStryker: Lesser beings don't see "red, yellow, and blue wool blocks". They see "wool-14,wool-4,wool-11" and might not make the connection. I annotated it here, but later when you do math examples it is the same, you mention different wood types which people might not know which blocks and meta values they represent.

AtomicStryker (approved):
Great changes! There was something i wanted to ask last time but forgot: What happens if you have grouped rules, lets say rule3=xxx with a variant ^xxx, and then a named rule variant rule4^xxx with MORE variants than the base rule? E.g. two variants ^xxx ^xxx under rule4.

Do you know the Ruins thread in minecraftforum? That's where the heavy users and content creators for Ruin templates reside - they will be doing most of the testing (if any) for your epic new system.

QuarterAnimal:
The first rule of the group calls all the shots. There's (currently) nothing wrong with another rule in the group having more variants than the leader, but those excess variants will never be chosen.

QuarterAnimal:
I presume you want them to make some noise. Now they do.

AtomicStryker:
Yes, more error messages can never be a bad thing (hahaha).
Now considering how to bring this to the users. I could merge this and simply push an update ... if there is a major issue, reverting is always possible ...

PR #141, opened 2017-09-10
versioning directives - allows template to work with any mod version

QuarterAnimal:
In the Forum, Gillymoth expressed an interest in building templates compatible with both old and new versions of the Ruins mod. I implemented some special comment directives to help with that. You can bracket a series of template comment lines with "#[[SHOW xxx.yy" and "#]]SHOW" to indicate those lines should be uncommented (i.e., the initial '#' is removed) when running with Ruins version xxx.yy or later. They'll appear as harmless comments to earlier versions. Conversely, "#[[HIDE xxx.yy" and "#]]HIDE" can be used to conceal legacy lines from Ruins version xxx.yy or later. Because, you know, things aren't confusing enough.

The change to RuinTemplate isn't quite as bad as it looks. Many of the redlines are just a consequence of removing layers of if-statements--comments are now absorbed by the new InputFilter class, so the rest of the code doesn't have to check for them.


QuarterAnimal:
A few small additions to help address the matter of unwanted air bubbles in underwater structures.

AtomicStryker:
Is there any reason to run older Ruins with "new" templates? I don't think there has been any point in time during which there was an advantage

QuarterAnimal:
To be honest, I wouldn't have thought so. Again, this was in response to Gillymoth's post in Minecraft Forum. I don't think it's a particularly good idea to try to make templates that support across a really wide span of versions, but it could be helpful during transitional times, especially when the mod changes break older templates--which, hopefully, isn't the situation here. I'm interested to see where the discussion in the Forum goes, if anyone makes a strong case for it. Otherwise, this isn't a feature I'm likely to ever use, myself. I just felt like coding something.

The bubbles...er...backgroundWater commit, though, _is_ something I'd use. Maybe I should put that on a separate fork?

AtomicStryker:
I would prefer that ... the versioning just adds complexity and stuff to maintain for no benefit i can think of.

And dare i say what Gillymoth actually wanted was probably a way to use variants in minecraft 1.10 and such

PR #142, opened 2017-09-11
backgroundAir/Water/Lava blocks, and expanded role of rule0

QuarterAnimal:
Just the background blocks.


AtomicStryker:
Add yourself to the developers while at it?

PR #143, opened 2017-09-12
background/foreground blocks and expanded role for rule0

QuarterAnimal:
With a little additional tweaking, I was able to extend the "background air" ("?air") concept to all blocks, and pick up the complementary "foreground" blocks almost for free. Background water ("?water") is still useful for eliminating air pockets around the base of submerged structures, but now you can also use other types of foreground and background blocks to fine-tune how a structure is seated in the surrounding environment. These are essentially the same mechanics currently used in site leveling--foreground blocks to fill holes and background blocks to flatten--made available for explicit use within rules.


QuarterAnimal:
...and, with this, I promise I'll be quiet for a while.

PR #146, opened 2017-09-17
Fix GUI overlay not being displayed.

bloodmc:
The wrong mappings were used in getEntityCrosshairOver during a recent update. This corrects the mappings and fixes the GUI overlay.


AtomicStryker:
Thank you. I did a quick test at some point and found it worked intermittently, so i wasn't sure why. Having the intersection box at the wrong coordinates would do it.

PR #150, opened 2017-10-09
reduce log spamming

QuarterAnimal:
I notice folks are cleverly using the /testruin command in command blocks, but it seems to have been designed with just testing in mind, not general use--it floods the log with all sorts of trace and debug messages which aren't particularly useful in a non-testing context. These changes squash the extraneous logging when the command wasn't issued by a player.


AtomicStryker:
Never even thought of that, although it seems obvious. God this code looks so magic and ugly by now.

PR #151, opened 2017-10-15
fix biomesToSpawnIn issues

QuarterAnimal:
The biomesToSpawnIn template parameter has some problems. It was being ignored, kind of (templates could only be added to biomes whose folder was already processed--so a template in "plains" could add itself to the "forest" spawn list, but not the list for "taiga"). Even when a template did manage to add itself to another biome's spawn list, the template counts and weights weren't updated. Now it should work as advertised...if not as expected, since this has probably been an issue for some time, and folks might've gotten used to the broken version.


AtomicStryker:
Thank you.

PR #152, opened 2017-10-16
recognize ALL the airs!

QuarterAnimal:
When used in a block rule, "minecraft:air" is flagged as a special block, then generates errors when it isn't recognized as such. The code is looking only for the explicit string "air" to define an air block; this modification allows it to treat "minecraft:air" as air, too.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
@@ -334,7 +334,7 @@ public class RuinTemplate
         RuinTemplateRule curRule;
 
         // height sanity check
-        final int y = Math.max(Math.min(yBase, world.getActualHeight() - height), 8);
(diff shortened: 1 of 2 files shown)


QuarterAnimal:
Three more issues I found while troubleshooting the community templates: substantial embed_into_distance values can push a structure below (or above) world limits, MediumMobSpawn and HardMobSpawn pseudoblocks are trying to spawn CaveSpiders instead of Cave_Spiders, and powered rail metadata is being rotated incorrectly.

AtomicStryker:
Have you checked whether or not the dynamic rotations file has mappings for the powered tracks? Hardcoding works, but is ugly

QuarterAnimal:
I did--it doesn't. Even if it did, the existing code wouldn't use it. If and when the code is refactored, I'd suggest using IBlockProperties.withRotation() and do away with both the hardcoded stuff AND the rotation mapping file.

AtomicStryker:
forgot this

PR #158, opened 2017-12-01
rotation reform

QuarterAnimal:
No new features this time, but a heck of a lot of cleanup, and (at least in theory) automatic support for blocks added by other mods. The rotation_mappings.txt file and CustomRotationMapping class are gone, as is a good chunk of messy block-rotation code in RuinTemplateRule. A number of existing rotation-related bugs were also squashed in the process--most notably some seriously botched teBlock rotation and placement, a particularly insidious issue with skull textures being rotated separately from the blocks themselves (turns out 0 isn't always a good default metadata value...), and assorted missing/broken cases.


AtomicStryker:
O_O

AtomicStryker:
I'm just going to merge this and look at it in eclipse, as it is too much to parse in the github diff view

QuarterAnimal:
Good luck! Here's the Big Picture:
- Metadata rotation was inconsistent from one block type to another. Some types guarded against rotation when oriented to NORTH, some didn't--and it isn't clear rotateMetadata() necessarily handled NORTH correctly, since it assumes it won't even be called. In an extreme case, the code for rotating and placing teBlocks was just flat-out wrong. I consolidated rotation and block instantiation into a single **realizeBlock()** function to help prevent inconsistency and eliminate a bunch of duplicate code.
- Tile entity rotation was not always happening when it needed to; specifically, for teBlocks and arbitrary IInventory containers. In fact, there was no support for tile entity rotation at all, aside from a special case for Skulls. Now there is; the cleverly named **rotateTileEntity()** function.
- There are API functions available for rotating arbitrary block states and tile entities, obviating the need for handling every individual as the special snowflake it is. I use them. I stripped the brobdingnagian switch statement out of RuinTemplateRule and **deleted rotation_mappings.txt and the CustomRotationMapping class**, as they're no longer needed. Ruins should now be able to rotate blocks like a boss, whether vanilla or modded, without having to maintain any sort of rotational instruction file.
- When a rule doesn't provide metadata for a block, 0 was assumed. For some block types (e.g., ladders, furnaces, skulls...), 0 isn't even a valid value, let alone a default. This led to some weird and puzzling effects, like rotating the texture of a skull without rotating the block itself. I now use the correct **default block state when no metadata is specified**, rather than 0.

I think this addresses all the block rotation issues. I tested it with both vanilla blocks and a few mod blocks, and it seems to work fine as far as I can tell. None of this directly affects existing templates at all; it does fix some bugs, though, so the new code may behave a bit differently than before in some situations. I did find a few existing rotation-related problems other than just blocks, but I'm looking at those separately. This changeset is dramatic enough already.

AtomicStryker:
Very nice! Now, please don't take this as me wanting to stop you (by all means continue), but keep in mind when you sink your time into this that Mojang may or may not make all of this moot when they manage to somehow completely break meta value based block saving for us.

QuarterAnimal:
I am considering that. While nothing is final, though, I'd bet these changes make Ruins significantly  _more_ robust with regard to the upcoming tsunami, not less. I'm using the newer, canonical class methods for block and tile entity rotation, and they should work--again, take this claim with a harvestcraft:saltItem--even after the 1.13 release. Note I'm not relying on metadata at all for the rotation itself, hence the elimination of rotateMetadata(); I'm working entirely with block states. Rotationwise, that is--RuinTemplateRule still uses metadata as its native state specification, so I need to support that.

Now, using metadata values in the _rule specifications_ is another matter altogether, but here I'm just addressing rotation. That was the gist of the Forum message I posted a few days ago--how does the template syntax evolve from "stone_button-2" to...um..."stone_button-face-wall-facing-west"? I have some ideas, but that's more of a decision for you to make, since any solution pretty much shatters any existing template files and kind of sets the direction the mod as a whole is going to take. I'd advocate adopting the standard command syntax (e.g., "stone_button[face=wall,facing=west]") and ditching the old "-metadata" scheme altogether, but that's your call.

Regardless of how they're specified, however, once blocks get into the mod, the rotation part should work just fine. One less thing to worry about when the storm hits.

AtomicStryker:
I was not aware there is a standard syntax yet (i am guessing Command Blocks use that?).

QuarterAnimal:
Commands in general, yes. You can see it in action by firing up the latest snapshot and using tab completion to walk you through, say, the /setblock command to place a fancy furnace: /setblock ~-1 ~ ~ furnace[facing=east,lit=true]{CustomName:"Fornax"} . (Watch out for bug MC-117574, though; replacing a block with another block of the same state currently fails, so you may have to destroy an old furnace to place a new one.)

PR #170, opened 2018-01-08
implement minimum distance from spawn parameter(s)

QuarterAnimal:
This adds a new **spawnMinDistance** template parameter, which specifies the minimum distance (Chebyshev distance, in blocks, on the XZ plane) from the world spawn point an instance of that template can be naturally generated in the overworld. If absent, the value defaults to 0 (i.e., no minimum distance is enforced). There's also a new **spawnMaxDistance** parameter, because...why not? Its default value is (effectively) infinity, which is also the value it gets if one attempts to set it to 0 (i.e., no maximum distance is enforced).

There are two new config parameters as well: **anySpawnMinDistance** and **anySpawnMaxDistance**. These specify the minimum and maximum distance limits for ALL template instantiation in the overworld, and will override any less-restrictive template specifications. The default value of anySpawnMinDistance is 32; the default value of anySpawnMaxDistance is infinity (no maximum limit). This mimics the mod's original behavior.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
@@ -37,6 +37,8 @@ class FileHandler
 
     public float templateInstancesMinDistance = 75f;
     float anyRuinsMinDistance = 0f;
+    public int anySpawnMinDistance = 32;
+    public int anySpawnMaxDistance = Integer.MAX_VALUE;
     static final HashSet<Block> registeredTEBlocks = new HashSet<>();
 
     private int templateCount;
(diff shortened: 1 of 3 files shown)


AtomicStryker:
I half-expected you to implement it lol, and there is no reason not to, maybe a general hint (don't take this as critical): Wait for the querying party to respond to your suggestion before implementing anything, they might want something different than you thought.

Business programming 101: THE CUSTOMER NEVER KNOWS WHAT HE WANTS

QuarterAnimal:
I'd like to add a lil sumpin sumpin to template_rules.txt documenting these new parameters, but I'm not sure whether or how you're planning to increment the version number. Will this be part of (the current) version 16.9, or the presumably upcoming 16.10? Or is it 17.0? 16.9.1? 16.9b? Looking back, it looks like you eschew two-digit "minor" version numbers, so I'd guess the answer is 17.0, but I don't want to assume.

AtomicStryker:
17.0 sounds good. Yeah i don't keep to a reasonable system. However, my changelogs actually contain all changes! :)

PR #195, opened 2018-04-27
[Ruins] remove incorrect pos file path validation (fixes #194)

### LINKED ISSUE #194: [Server] Ruins can fail to properly use level-name from server.properties
MC: 1.7.10
Ruins: 1.7.10

__Issue:__

When loading a server, Ruins can fail to properly parse 'level-name' in 'server.properties' if 'level-name' includes a path. This results in Ruins not being able to locate 'RuinsPositionsFile.txt' and not generate any ruins in game.

__Details:__

Minecraft allows server admins are to specify a path for 'level-name' in server.properties if they want to keep all their multiplayer worlds in a single folder (similar to how single-player does):

*__level-name=/Worlds/MyFunWorld__*

The above would create a subfolder called 'Worlds' in the server root and then place the level folder 'MyFunWorld' inside it. The problem is, Ruins is expecting a simple world name without a path:

*__level-name=MyFunWorld__*

as a result, when including a path in 'level-name', Ruins generates an error and fails to function:

*__Ruins attempted to load invalid worldname /Worlds/MyFunWorld posfile__*

__NOTE:__

*I originally posted this on someone else's fork by accident...*

I'm not sure if you're supporting 1.7.10 however the reason I have not moved from that version is because I also have the LotR mod which has not updated beyond that. It would certainly be appreciated if you had the time to fix this one issue for 1.7.10,

Thanks!

QuarterAnimal:
An attempt is made to validate the position data file path by checking to make sure it contains the world name string. This validation fails on Windows servers if the world name is a path, though, since Windows uses backslashes to separate path elements instead of Java's native slashes. One solution would be to convert the path to a file system-neutral expression, but given how the path is generated, this validation is unnecessary and can be removed altogether, avoiding the separator issue completely.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinGenerator.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinGenerator.java
@@ -41,14 +41,7 @@ class RuinGenerator
         ruinsDataFile = new File(rh.saveFolder, fileName);
         ruinsDataFileWriting = new File(rh.saveFolder, fileName + "_writing");
 
-        if (ruinsDataFile.getAbsolutePath().contains(world.getWorldInfo().getWorldName()))
-        {
-            new LoadThread().start();
-        }
-        else
-        {
-            System.err.println("Ruins attempted to load invalid worldname " + world.getWorldInfo().getWorldName() + " posfile");
-        }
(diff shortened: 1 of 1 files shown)


AtomicStryker:
Thank you.

PR #197, opened 2018-05-03
[Ruins] remove embed distance from surface search

QuarterAnimal:
A template's embed_into_distance shouldn't factor into the search for an acceptable surface, since the structure is to be embedded into that surface _after_ one is found. Without this fix, structures are dragged down further than they should be, possibly low enough so that no surface can be found. This is the root cause behind Gillymoth's swamp structures not spawning when they should, as mentioned in [Forum post 4349](https://www.minecraftforum.net/forums/mapping-and-modding-java-edition/minecraft-mods/1282339-1-12-ruins-structure-spawning-system?comment=4349).

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
@@ -199,10 +199,10 @@ public class RuinTemplate
         }
 
         // guess the top Y coordinate of the structure box, for checking top to bottom
-        final int topYguess = y + height - embed + additionalYRangeChecked;
+        final int topYguess = y + height + additionalYRangeChecked;
 
         // set a lowest height value at which surface search is aborted
-        final int minimalCheckedY = y - height - embed - additionalYRangeChecked;
+        final int minimalCheckedY = y - height - additionalYRangeChecked;
 
         // surface heights of the proposed site, -1 means 'out of range, consider overhang'
         final int[][] heightMap = new int[xDim][zDim];


AtomicStryker:
I'll just trust you and merge this, in my head there's questions though: If this happens to all templates, why is a problem specifically in swamps? Does this hurt template spawning chances overall, as the "guessed" spawning coordinates are now overall higher than before and terrain appears more un-even than before?

QuarterAnimal:
It's not a swamp-specific problem, it's just that certain of Gillymoth's swamp templates happened to meet the conditions where the problem becomes clearly manifest--namely, an **embed_into_distance** greater than the sum of the structure height and the depth of acceptable target blocks.

In the case of Gillymoth's fishing village, for example, **embed_into_distance** is 10, the template height is 3 (it's a 3x3x3 cube of command blocks), and water--the acceptable target medium--in swamps is typically shallow enough (say, 4 or 5 blocks) so that the search for a surface (namely, **checkArea()**) is completely buried in the ground beneath the water. Thus no surface is found, **checkArea()** fails, and the structure is inhibited.

Actually, you can see trouble brewing even before a structure is prevented from spawning altogether. If **embed_into_distance** is greater than the template height, or less than the negative of the template height, the excess embedding amount is _doubled_. So, in the fishing village example above, if you set **embed_into_distance** to 3 (the template height), the structure spawns just under the water surface, as expected. But if you set it to 4, it spawns _2_ blocks lower, not 1. Set it to 5, and it spawns 4 blocks lower. And so on.

The change only affects templates in that trouble zone. Templates designed to spawn structures on or at least partly above the surface--and not hovering far above it--will spawn exactly as before. Buried (or high-flying) structures won't be buried (or flying) quite as much; they'll now be embedded by the correct amount, without the bonus doubling effect. And some templates at the extremes that couldn't be instantiated at all, like Gillymoth's fishing village, now will be.

I'm not sure I follow the concern about higher coordinates and more uneven terrain. I don't think either of those is the case. The basic idea is simply to put **checkArea()** on the same page with regard to embedding as **findSuitableY()** is--namely, they both ignore it. What happens without this fix is **findSuitableY()** looks for an acceptable surface, then **checkArea()** effectively looks for an embedded surface _below_ that surface...and if it's embedded far enough, it won't see the original surface **findSuitableY()** found.

QuarterAnimal:
Upon further testing, I discovered the effect of this issue on "flying" templates (i.e., with negative **embed_into_distance** values) is rather different from what I described above--I was focused on entirely buried templates like the fishing village. Turns out flying templates don't get the double-embedding boost that buried ones do; instead, at some point, they become completely blind to the terrain and count everything as overhang. If template authors (incorrectly, it could be argued) set a full-footprint **allowable_overhang** for fliers, the error is masked and has no discernible effect. If the overhang _is_ properly constrained, however, the template won't spawn at all without this fix.

PR #212, opened 2018-07-07
[Ruins] generalize comma hiding for rule parsing

QuarterAnimal:
Ruins currently has a method for substituting {brace-delimited NBT tags} in a rule with placeholders so as to hide commas that'd screw up its simple split-on-comma parsing, and another to restore the original text. Going forward, [bracket-delimited block state properties] may also contain commas, so this pull request anticipates upcoming changes by replacing the NBT tag substitution and restoration methods with a more general-purpose utility. It also addresses some shortcomings in the current code, such as its failure to properly account for quoted strings containing "false" delimiter characters and placeholder-like text which could confound the restoration process.

While this pull is primarily intended as part of the preparation for Minecraft 1.13, it isn't at all 1.13-specific and works just fine in 1.12.


AtomicStryker:
Please refactor the Lumper into its own java file

PR #213, opened 2018-07-08
[Ruins] biomes to spawn in by type

QuarterAnimal:
This change allows a template to alternatively (or additionally) specify the biomes in which it can be instantiated by Forge's BiomeDictionary.Type instead of name. So, instead of explicitly listing all forest-like biomes...
**biomesToSpawnIn=forest,birch_forest,taiga,roofed_forest,forest_hills,yada,yada,yada**
...one can simply specify one or more types...
**biomeTypesToSpawnIn=forest**
This will also catch any biome--vanilla or modded--matching at least one of the specified types. The special type "all" can be used to indicate _all_ biomes, regardless of type (even including biomes with no type).

To further fine-tune the list of eligible biomes, the new variables **biomesToNotSpawnIn** and **biomeTypesToNotSpawnIn** are available, allowing the exclusion of biomes by name or type, respectively. Note these only affect biomes included via **biomeTypesToSpawnIn**; biomes listed as **biomesToSpawnIn** are _always_ candidates for spawing--as is the biome corresponding to the template's folder, of course--even if they appear on the exclusion lists. Here's a more comprehensive example:
**biomesToSpawnIn=mushroom_island
biomeTypesToSpawnIn=forest,jungle
biomesToNotSpawnIn=birch_forest
biomeTypesToNotSpawnIn=hills,mountain**
A template with these variable settings can spawn on mushroom islands (but not mushroom island shores, a different biome) and in any biome that's either a forest or jungle type (or both), _unless_ it's also a birch forest, or a hill or mountain type. Whew!

I expect in the real world, the typical case will be to just specify **biomeTypesToSpawnIn** to apply a particular "theme" to a template. This avoids the need for lengthy **biomesToSpawnIn** lists, command block proxies and template duplication, and the omission of unanticipated biomes added by other mods.

The default biome types are (currently): hot, cold, sparse, dense, wet, dry, savanna, coniferous, jungle, spooky, dead, lush, nether, end, mushroom, magical, rare, ocean, river, water, mesa, forest, plains, mountain, hills, swamp, sandy, snowy, wasteland, beach, and void. As mentioned above, an additional special type--all--may be used for **biomeTypesToSpawnIn**. Mods can also create new biome types, so this list isn't necessarily complete, but Ruins will fully support modded types as well.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
@@ -9,6 +9,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Random;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -20,6 +21,8 @@ import net.minecraft.init.Blocks;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
+import net.minecraft.world.biome.Biome;
+import net.minecraftforge.common.BiomeDictionary;
 import net.minecraftforge.common.IPlantable;
 import net.minecraftforge.common.IShearable;
 import net.minecraftforge.common.MinecraftForge;
@@ -778,6 +781,10 @@ public class RuinTemplate
 
     private void parseVariables(ArrayList<String> variables) throws Exception
     {
+        Set<String> included_biomes = new HashSet<>();
+        Set<String> included_biome_types = new HashSet<>();
+        Set<String> excluded_biomes = new HashSet<>();
+        Set<String> excluded_biome_types = new HashSet<>();
         Iterator<String> i = variables.iterator();
         String line;
         while (i.hasNext())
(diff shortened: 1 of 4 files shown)


AtomicStryker:
Is there a particular reason you "normalize" biome types by uppercasing but not the biome names?

QuarterAnimal:
I do that for biome types because BiomeDictionary does that, too. Internally, biome type names are all uppercase; however, the getType() method for converting a type name to a type uses toUppercase() on the given name before looking it up, so they are case insensitive by design.

I didn't do it for biome names because it wasn't done before. Biome names are typically lowercase by convention, but I don't know that that's enforced by the code. I suspect not, which would mean biome names are technically case sensitive and shouldn't be altered (to avoid the risk of breaking compatibility with other, especially older, mods).

MentalMouse:
Hmm.  How would I combine two selectors, e.g. if I wanted "dry forests"? 

BTW, I note that your biome types in your example are not in fact uppercased.  I like having the biomes and the types distinguished from each other.

QuarterAnimal:
You wouldn't. You could specify all dry biomes and all forest biomes combined, but there's no way to select only biomes that are both dry and forest. If you really think that's vital functionality, I could whip something up, but what's the tradeoff between capability and complexity? If you can do dry&&forest, are you going to also expect to be able to do (dry||hot)&&(forest||(rare&&!plains))?

As for case, uppercase type names work just as well as lower, if that's the template author's preference. It's not mine, personally. The type names are internally uppercase only for legacy reasons; Minecraft has since converged on a standard naming convention of all lowercase for pretty much everything (though NBT tag names are still adorably inconsistent). Still, I see your point about ease of differentiation between names and types, and there's nothing stopping you from using
**biomeTypesToSpawnIn=FOREST,JUNGLE**
in a template--it'll still fly. Or are you suggesting I should force people to use uppercase?

AtomicStryker:
Nono, just curiosity. And a vague memory of upper/lowercase problems with different platforms (damn you, windows) but i think those stopped when MC went all lowercase on biome names.

MentalMouse:
QuarterAnimal: Depends what you consider "vital", but I'd consider that useful, even the complex expressions.  I'd go so far as to say it would make the difference between "crude control" and "fine control" of biome selection.

QuarterAnimal:
Okay, we'll start simple (see pull request #214). I think I'd like to see how folks actually use this feature before hanging all sorts of bells and whistles on it.

MentalMouse:
Thanks!  That does seem to cover the basic needs for biome selection, with the comma list providing for a top-level OR.

PR #214, opened 2018-07-08
[Ruins] add simple boolean operations to biomeTypesToSpawnIn

QuarterAnimal:
Stupid ENTER key jumped right out in front of me before I could type a comment! I swear!

This change addresses MentalMouse's comments on pull request #213 by allowing **biomeTypesToSpawnIn** to accept rudimentary boolean operations in addition to just plain type names. Type names still work as before--they're essentially degenerate cases. But now you can concatenate multiple type names together, separated by + (AND) operator, to specify a biome that has _all_ those types. So, if you want to spawn only in cold forests or snowy mountains, you can set:
**biomeTypesToSpawnIn=cold+forest,snowy+mountain**
This means in order to be a spawn candidate, a biome must be both cold AND forest types, or both snowy AND mountain types. There's also a - (AND NOT) operator:
**biomeTypesToSpawnIn=hot+dry-nether**
This means a biome must be both hot AND dry, AND NOT nether. These operators may also be used in specifying **biomeTypesToNotSpawnIn**, for those who can follow the twisty logic.

### DIFF
--- a/Ruins/src/main/resources/examplesAndDocs/template_rules.txt
+++ b/Ruins/src/main/resources/examplesAndDocs/template_rules.txt
@@ -170,6 +170,16 @@
 # example: biomeTypesToNotSpawnIn=spooky,rare
 # Optional. Only relevant if biomeTypesToSpawnIn list is not empty.
 #
+# In both the biomeTypesToSpawnIn and biomeTypesToNotSpawnIn lists, some rudimentary boolean
+# operations are allowed by concatenating multiple type names separated by + (and) or - (and
+# not). So, for example:
+# biomeTypesToSpawnIn=forest-cold,jungle+hills
+# This adds the template to all biomes that are of type forest AND NOT cold, or are of both
+# jungle AND hills. Order doesn't matter, so the following is equivalent:
+# biomeTypesToSpawnIn=hills+jungle,-cold+forest
+# If all the types are exclusions, an implicit ALL is added. Thus, "-nether-end" is the same
+# as "all-nether-end" (i.e., any biome that isn't of type nether or end).
+#
 
 weight=5
 embed_into_distance=1
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinBiomeTypeCriteria.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinBiomeTypeCriteria.java
@@ -1,0 +1,123 @@
+package atomicstryker.ruins.common;
+
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import net.minecraft.world.biome.Biome;
+import net.minecraftforge.common.BiomeDictionary;
+
+// a biome type criteria collection against which biomes are checked
+// as to whether or not they satisfy at least one of the specified
+// conditions
+class RuinBiomeTypeCriteria
+{
+    private RuinTemplate template_;
+    private PrintWriter log_;
+
+    // a biome satisfies a particular criterion if it is assigned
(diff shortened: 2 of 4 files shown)


MentalMouse:
Looks good, thanks!

Ruins/src/main/resources/examplesAndDocs/template_rules.txt
```diff
@@ -170,6 +170,16 @@
 # example: biomeTypesToNotSpawnIn=spooky,rare
 # Optional. Only relevant if biomeTypesToSpawnIn list is not empty.
 #
+# In both the biomeTypesToSpawnIn and biomeTypesToNotSpawnIn lists, some rudimentary boolean
+# operations are allowed by concatenating multiple type names separated by + (and) or - (and
+# not). So, for example:
+# biomeTypesToSpawnIn=forest-cold,jungle+hills
+# This adds the template to all biomes that are of type forest AND NOT cold, or are of both
+# jungle AND hills. Order doesn't matter, so the following is equivalent:
+# biomeTypesToSpawnIn=hills+jungle,-cold+forest
```
AtomicStryker, line 179: -cold? typo? Or wait, i see, the + is implicit above

AtomicStryker:
Please refactor BiomeTypeCriteria into its own java file, it is big enough and has self-contained logic.

Not a huge fan of this, you basically need an understanding of logic operations to even use it, there is multiple "hidden"/implicit mechanics... but the 3 people who are going to use this might need it just so ...

QuarterAnimal:
I don't entirely disagree. I tried to keep it simple, though, and somewhat intuitive for anyone with a basic grasp of programming or algebra--the notion of "adding" and "subtracting" acceptance criteria. I also wanted to make it unobtrusive so users could still simply list biome types if they didn't need to deal with the more complicated stuff.

If you'd like, I can get rid of some or both of the implicit elements:
1) The implicit + in front of the first type if no other operator is present. I see this as an ease-of-use feature--it's what allows you to just provide a list of types as before. If I require the first operator to be present, the +- business is in your face whether you want to use it or not. Note you _can_ explicitly put an initial + already if you want to; it's optional, but legal:
**biomeTypesToSpawnIn=+cold+forest,+snowy+mountain**
I think that's a bit tougher to read myself, with that sneaky, easy to miss comma. On the other hand, it could be argued it's more consistent: every type is preceded by an operator, always. Should that be mandatory?
2) The implicit +ALL added to a criteria specification in which every constituent criterion is exclusive. In other words, "-nether-end" is interpreted as "all-nether-end"; otherwise, "-nether-end" doesn't _include_ anything and won't match any biome. I suspect most people would think "-mountain" to mean "anything that isn't a mountain" rather than "nothing, but especially not a mountain," but maybe that's just me. Again, you can explicitly specify +ALL if you want to. Should that be mandatory?

AtomicStryker:
Actually, what happens if a biometype has a hyphen in it? "volcano-cauldron"?

QuarterAnimal:
Then somebody needs to die. According to the API documentation for BiomeDictionary, a type name must not contain special characters. I'm not accounting for that possibility any more than, say, rule parsing accounts for the possibility of block names containing commas. Sure, it _could_ happen, but...

It should be noted the creation of custom biome tags is rare, and for good reason: it kind of defeats the purpose of having a reliably standard set of biome descriptors. Of all the mods I looked at, I only found one that does it--Recurrent Complex--and it follows the no-special-character rule (it also only applies its new types to vanilla biomes, significantly limiting their utility, so I'm not really sure what the point of them is).

However, if it's a real concern, I could add support for escaping characters, or enclosing type names in quotes. I'm reluctant to do that, though, because we're nearing the point where every single line in the template file has its own unique grammar ("why can I escape a hyphen in a biome type, but not a comma in a block name?"). I'd _love_ to refactor all parsing throughout the mod--there's plenty of room for improvement, and that was the motivation for my raising the discussion of standardizing names in the Forum--but I'm trying to abide by your "change as little as needed" dictum.

PR #216, opened 2018-07-09
[Ruins] templates can declare required/prohibited mods

QuarterAnimal:
A feature I have in my local experimental Ruins and requested by Jordan_Peacock in a [Minecraft Forum post](https://www.minecraftforum.net/forums/mapping-and-modding-java-edition/minecraft-mods/1282339-1-12-ruins-structure-spawning-system?comment=4413)--the ability for a template to declare mods that must be present, or can't be present, for it to be loaded. This is done via two new template variables: **requiredMods** and **prohibitedMods**. Pretty self explanatory. They both default to empty lists (i.e., nothing required, nothing prohibited).

So, for example, if I have a template that relies on blocks from the Quark and Tinkers' Construct mods, I'd specify:
**requiredMods=quark,tconstruct**
Then maybe I have an alternative template that only uses Quark:
**requiredMods=quark
prohibitedMods=tconstruct**
And one that only uses Tinkers':
**requiredMods=tconstruct
prohibitedMods=quark**
And finally one that doesn't use either:
**prohibitedMods=quark,tconstruct**
Only one of these four templates will be loaded by Ruins, depending on the combination of other mods installed. I don't expect people will use it that way, but they could. It's more likely they'll simply use **requiredMods** as a guarantee everything the template needs is available. That's cool, too.

Jordan_Peacock suggested it'd be ideal if Ruins made the determination itself, based on analyzing the contents of the template file. I went the variable route, because:
- The analysis would be quite complex, given that not only modded blocks in rules would need to be checked, but also entities, blocks, items, effects, etc. embedded in NBT tags and referred to in commands and inventories. Then there are modded biomes; Ruins doesn't support qualifying them with their associated mod names. Messy, messy.
- You may want to require a mod for reasons other than additional things it provides. Maybe, for example, I want some extra iron deposits sprinkled around to feed Thermal Expansion, but I don't want the excess iron if there's no big consumer of it. Nothing in the template file would indicate this.
- There's also no way to determine by the contents of a template whether a particular mod is _prohibited_, if that's ever a thing.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
@@ -25,6 +25,7 @@ import net.minecraft.world.biome.Biome;
 import net.minecraftforge.common.IPlantable;
 import net.minecraftforge.common.IShearable;
 import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.fml.common.Loader;
 
 public class RuinTemplate
 {
@@ -778,8 +779,17 @@ public class RuinTemplate
         }
     }
 
+    private static Set<String> active_mods_ = null;
+
     private void parseVariables(ArrayList<String> variables) throws Exception
     {
+        // collect list of currently active mods on first pass and keep for future reference
+        if (active_mods_ == null)
+        {
+            active_mods_ = new HashSet<>();
+            Loader.instance().getActiveModList().forEach(mod -> active_mods_.add(mod.getModId()));
+        }
+
         Set<String> included_biomes = new HashSet<>();
         RuinBiomeTypeCriteria included_biome_types = new RuinBiomeTypeCriteria(this, debugging ? debugPrinter : null);
         Set<String> excluded_biomes = new HashSet<>();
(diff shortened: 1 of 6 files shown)


Ruins/src/main/java/atomicstryker/ruins/common/World2TemplateParser.java
```diff
@@ -459,6 +459,9 @@ private void toFile(File file)
             pw.println("# authoring Player: " + player.getName());
             pw.println();
 
+            pw.println("# other mods required or prohibited by this template");
```
AtomicStryker: Please specify "list of mod id seperated by comma"

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
```diff
@@ -1248,4 +1310,34 @@ public RuinTemplateRule getVariant(final int selectorInitial, final int ruleInde
             }
         }
     }
+
+    public static class ClassException extends RuntimeException
+    {
+        private static final long serialVersionUID = -7630819126112904714L;
+
+        protected ClassException(String message)
```
AtomicStryker: ClassException is a bad name, TemplateLoadingException maybe? Also, adding three classes for 2 errors which are actually somewhat the same is over-engineering things, IMO

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
```diff
@@ -1248,4 +1310,34 @@ public RuinTemplateRule getVariant(final int selectorInitial, final int ruleInde
...
         }
     }
+
+    public static class ClassException extends RuntimeException
+    {
+        private static final long serialVersionUID = -7630819126112904714L;
+
+        protected ClassException(String message)
+        {
+            super("RuinTemplate: " + message);
```
AtomicStryker: Resolves to "RuinTemplate: template derp..."
Redundant information, and i would ask you to refactor those 3 exceptions into a single type anyway

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
QuarterAnimal: I want something a little more specific, as callers may want to distinguish between this particular situation and something else. Other exceptions may indicate more serious issues and warrant corrective action or urgent log messages; this one is pretty benign and could very well be intentional/expected. I went with IncompatibleModException.

QuarterAnimal:
I squashed a stray ALL biome type lurking in the boilerplate template comments, too

PR #218, opened 2018-07-11
[Ruins] cache block states for efficient repeat use

QuarterAnimal:
This, like the comma-hiding pull #212, is another change to prepare for Minecraft 1.13, though it too isn't 1.13-specific and still works like a charm with 1.12. In fact, it provides a marginal improvement in 1.12, with (potentially) quicker structure-building performance and more thorough error reporting.

The main feature is the caching of computed block states rather than reinterpreting the data values each time a block is placed. It's not that significant a gain in 1.12, but it'll be much more relevant in 1.13, when parsing property lists will be kind of expensive. This change is transparent to end users (aside from now logging bad block data values in templates).

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
@@ -49,7 +49,8 @@ import net.minecraftforge.fml.relauncher.ReflectionHelper;
 public class RuinTemplateRule
 {
     protected final Block[] blockIDs;
-    protected final int[] blockMDs;
+    protected final String[] blockMDs;
+    protected final IBlockState[] blockBSs;
     protected final String[] blockStrings;
     protected final SpecialFlags[] specialFlags;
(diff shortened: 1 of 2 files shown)


Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
```diff
@@ -1461,6 +1457,51 @@ private static TileEntity realizeBlock(World world, int x, int y, int z, Block b
...
+        if (state == null)
+        {
+            // convert data value to state
+            String metadata = blockMDs[blocknum];
+            if (metadata != null && !metadata.isEmpty())
+            {
+                try
+                {
+                    int value = Integer.parseInt(metadata);
+                    if ((value & ~0b1111) == 0)
```
AtomicStryker: document this bit filter for mere mortals

QuarterAnimal: Ha! Knew you'd like that.

PR #220, opened 2018-07-15
[Ruins] boolean expression refinement & generalization

QuarterAnimal:
So, somebody asked me "why can't I use the same kinds of boolean expressions for **requiredMods** that I can for **biomeTypesToSpawnIn**?" I didn't have a good answer. If you want an iron-ore laden template to spawn if you have any one of five iron-hungry mods loaded, there's no way to do that. You can specify all five together are required, but you can't specify any subset of the five.

At least until now. I've generalized the boolean expression parser to be used for both parameters, and maybe more in the future. If you want to make a template dependent on all three mods alpha_mod, beta_mod, and gamma_mod, specify:
**requiredMods=alpha_mod+beta_mod+gamma_mod**
Read this as "alpha_mod AND beta_mod AND gamma_mod." If the template should spawn if any one of the three is present, that's:
**requiredMods=alpha_mod,beta_mod,gamma_mod**
which reads as "alpha_mod OR beta_mod OR gamma_mod." There's no need for **prohibitedMods** any more, since that can be specified with the same parameter:
**requiredMods=red+blue-yellow-green**
meaning "red AND blue AND NOT yellow AND NOT green." The **biomeTypesToNotSpawnIn** parameter goes away, too, since it's unnecessary for the same reason.

I also now support quoted strings, complete with escaped sequences, since legacy mod names might contain hyphens and other crappy stuff:
**requiredMods="Old-Hyphenated-Mod"+"))wacko, crazyname(("+"##\\\"qostr\"\\###"**
That carries over to biome types as well. Just to be safe, I removed the forced case-casting, so biome types must match their names as documented (i.e., uppercase):
**biomeTypesToSpawnIn=RARE+(WET,HOT)**
meaning rare, and either wet or hot. Oh, yeah...parentheses now, too.

I know this adds some complexity, but it's now fully-featured, and I thought if it's something to be considered, it should go in before the previous changes go live. Code-wise, it's not that much more complicated than before; it just looks bigger because I added a crapton of comments and diagnostic messages.


Ruins/src/main/java/atomicstryker/ruins/common/RuinVennCriterion.java
```diff
@@ -0,0 +1,656 @@
...
+ * anything following it (unless enclosed in quotes).
+ * <p>
+ * Example: A criterion created from the expression <code>RED + (YELLOW, GREEN) - BLUE</code> is satisfied by any
+ * collection of strings containing RED, either TELLOW or GREEN (or both), but not BLUE.
+ * <p>
+ * Optionally, a collection of elements may also be provided from which an expression must be constructed. If an
+ * element in the expression does not appear in the list of valid elements, an exception is thrown (if strict
+ * validation is enabled) or a warning is logged (if strict validation is not enabled).
+ */
+class RuinVennCriterion
```
AtomicStryker, line 44: Why "Ruin", for starters it would be "Ruins" but also there is nothing in here specific to Ruins

AtomicStryker:
Thats is one heap of math. I'm sorry i have to do this to you, i wish there was another choice, but...
I will require ... UNIT TESTS.

Ruins/src/main/java/atomicstryker/ruins/common/RuinVennCriterion.java
QuarterAnimal: Half the files start with "Ruin"...I thought that was your naming convention. "Ruins" only appears in one file name: RuinsMod. I'm happy to change it if you'd like.

AtomicStryker: There's a difference between a Ruin and Ruins, the mod :)

AtomicStryker: Not to say i'm not inconsistent, because i probably am

PR #224, opened 2018-07-25
[Ruins] implement named block properties

QuarterAnimal:
This is the biggie, replacing the -<data_value> syntax for specifying block metadata in a template rule with [<property_name>=<value>,...]. So, instead of:
```
rule1=0,100,gravel
rule2=0,100,wool-6
rule3=0,100,log-7
rule4=0,100,sapling-2-addbonemeal
```
templates should use:
```
rule1=0,100,gravel
rule2=0,100,wool[color=pink]
rule3=0,100,log[axis=x,variant=jungle]
rule4=0,100,sapling[type=birch]-addbonemeal
```
Because it may be needed to support some older mods, data values can still be used with the new syntax:
```
rule1=0,100,gravel
rule2=0,100,wool[6]
rule3=0,100,log[7]
rule4=0,100,sapling[2]-addbonemeal
```
This is discouraged, though, and should be considered a deprecated feature.

Also changed to accommodate property lists: templates generated by the **/parseruin** command, block data displayed by the query stick, and the default templates. This is completely functional in Minecraft 1.12; the templates will have to be modified again when Forge for Minecraft 1.13 becomes available, along with some (hopefully minor) code changes, due to the new block IDs and properties.


MentalMouse:
Are you making a conversion tool?  Perhaps a change in file extension would be warranted too.

QuarterAnimal:
I wasn't planning on making a conversion tool. Certainly not for 1.12, since everything would just have to be converted _again_ in a couple weeks (?) to be compatible with all the new block names in 1.13. Once there's a Forge for 1.13, I'll see what the API provides and consider it...but even in the best case, automating conversion is problematic. Depending on what's exposed, it might be fairly simple to throw something together that does about 80% of the work, but there's definitely stuff that has to be done manually anyway--modded blocks and items, NBT tags, commands, new data value<->property mappings. I don't know that it's worth it.

AtomicStryker:
But WHY would you implement this before 1.13 is out
I am not merging this to a 1.12 branch

AtomicStryker:
As for conversion: The ingame parser will have to do. If people have custom templates, they can spawn them in a 1.12 save, go to 1.13, and re-parse them. As some other things will have to go entirely (such as the rotation system) i see little alternative.

QuarterAnimal:
I wouldn't expect it to go into a 1.12 _release_. I presume you'll eventually be spinning off a 1.13 branch at some point, though.

It does work in 1.12, and you can play around with it now if you're so inclined. I'm not sure I understand your concern regarding rotation, for example, because I see it working just fine. When Forge for 1.13 comes out, there should only be a few small changes needed to this code. The templates are another story.

AtomicStryker:
Because while this code is neat in theory, you don't know if you will have to redo half for the actual 1.13 realities? I see you build a lot using "hardcoded" Stringbuilders and patterns, what if vanilla has a convenience method to turn blocks into strings? All that effort, possibly wasted

From what i've read, some of the new flat blocks like fences store their connections(!) to other blocks. I think torches now also store their headings. Unless vanilla provides a standard rotation API, this will mean templates with these blocks will break when rotated.

See this https://gist.github.com/williewillus/353c872bcf1a6ace9921189f6100d09a

QuarterAnimal:
Vanilla always has had convenience methods for turning blocks into strings. Ruins even uses them in its various diagnostic messages. The StringBuilders and Patterns I use are for reading and writing template files; I feel pretty confident neither Mojang nor Forge is ever going to provide methods for translating to and from Ruins template syntax.

Vanilla does provide a standard rotation API, and Ruins has been using it since release 17.0 in January. Minecraft uses it for its own structure mechanics, so it's bound to exist in some form. It's more than capable of readily handling fence connectivity as a server-side property; it already works with vines, for example, which has a similar multiface aspect today.

QuarterAnimal:
To be revisited in 1.13.x...

PR #228, opened 2018-09-25
Dungeon Tweaks Compat again

jredfox:
this has backwards compatibility with older dungeon tweaks and the core function re-write update of 1.2.5. This code is also cleaner I hope you will merge this. **And yes still this is still a soft dependency**

Why do you need to register dungeon entries by default now?: it's to instantiate weights that are not 0 for towers. Users are still able to override the weights via config file it's just the weight won't be default. Also my re-write didn't support any specific mob entries only definitions. If you need the method to be taken down I could add support on my side for weights but, the rest needs to happen.

Also if you don't like my nether specific enhancements by default when dungeon tweaks is loaded I could remove that and have users config them to be that or whatever they wanted. Just thought it would be cool. 

Thanks for your time

### DIFF
--- a/BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
+++ b/BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
@@ -82,9 +82,9 @@ public class DungeonTweaksCompat {
                     world);
 				MinecraftForge.EVENT_BUS.post(event);
 			}
-			catch (Throwable t)
+			catch (Exception e)
 			{
-				t.printStackTrace();
+				e.printStackTrace();
 			}
 		}
 		else
@@ -94,9 +94,9 @@ public class DungeonTweaksCompat {
 				Method fireDungeonTweaks = Class.forName("com.evilnotch.dungeontweeks.main.world.worldgen.mobs.DungeonMobs").getMethod("fireDungeonTweaks",ResourceLocation.class,TileEntity.class,Random.class,World.class);
 				fireDungeonTweaks.invoke(null, towerId, spawner, random, world);
 			}
-			catch(Throwable t)
+			catch(Exception e)
 			{
-				t.printStackTrace();
+				e.printStackTrace();
 			}
 		}
 	}
--- a/BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
+++ b/BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
@@ -1,0 +1,104 @@
+package atomicstryker.battletowers.common;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.util.Random;
+
+import com.evilnotch.dungeontweeks.main.world.worldgen.mobs.DungeonLocation;
+import com.evilnotch.dungeontweeks.main.world.worldgen.mobs.DungeonMobs;
+
+import atomicstryker.battletowers.common.AS_WorldGenTower.TowerTypes;
+import net.minecraft.tileentity.TileEntity;
+import net.minecraft.tileentity.TileEntityMobSpawner;
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.fml.common.eventhandler.Event;
+/**
+ * this is backwards and modern dungeon tweaks compatible
+ * @author jredfox
+ *
+ */
+public class DungeonTweaksCompat {
+	
+	public static boolean isLegacy = false;
+	
+	/**
+	 * make backwards compatability when isLegacy becomes true
+	 */
+	public static void legacyCheck()
+	{
+		try
+		{
+			Class c = Class.forName("com.EvilNotch.dungeontweeks.main.Events.EventDungeon$Post");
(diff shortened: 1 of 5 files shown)


BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
```diff
@@ -0,0 +1,104 @@
...
+		{
+			
+		}
+	}
+	/**
+	 * register all dungeon tweaks mobs to anydim towers
+	 */
+	public static void registerDungeons()
+	{
+		if(isLegacy) return;//I supported this mod in older versions
```
AtomicStryker: Formatting

BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
```diff
@@ -0,0 +1,104 @@
...
+			Method addDungeonMob = Class.forName("com.evilnotch.dungeontweeks.main.world.worldgen.mobs.DungeonMobs").getMethod("addDungeonMob", ResourceLocation.class,ResourceLocation.class,int.class);
+			for(AS_WorldGenTower.TowerTypes tower : AS_WorldGenTower.TowerTypes.values())
+			{
+				boolean nether = tower == TowerTypes.Netherrack;
+				addDungeonMob.invoke(null, tower.getId(), new ResourceLocation("cave_spider"), 100);
+				addDungeonMob.invoke(null,tower.getId(), new ResourceLocation("spider"), 90);
+				addDungeonMob.invoke(null,tower.getId(), nether ? new ResourceLocation("wither_skeleton") : new ResourceLocation("skeleton"), 120);
+				addDungeonMob.invoke(null,tower.getId(), new ResourceLocation("zombie"), 120);
+
+				if(nether)
```
AtomicStryker: formatting

BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
```diff
@@ -0,0 +1,104 @@
+package atomicstryker.battletowers.common;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.util.Random;
+
+import com.evilnotch.dungeontweeks.main.world.worldgen.mobs.DungeonLocation;
+import com.evilnotch.dungeontweeks.main.world.worldgen.mobs.DungeonMobs;
```
AtomicStryker: Pretty sure the idea of compat proxies is to avoid having to IMPORT the addons?

AtomicStryker (commented):
How is this a soft dependency if there are flat out import statements? Unless they are API classes you added? Or do i misunderstand something?

BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
jredfox: ok I will remove the imports but, I don't use them. I converted them to reflection so you don't have to have my mod to compile even though it would be safe using it without the mod since loader is mod loaded

BattleTowers/src/main/java/atomicstryker/battletowers/common/DungeonTweaksCompat.java
jredfox: yes I will download your formattor again I guess

jredfox:
import statements do not exist in compiled code. **I removed those import statements** and the dungeon tweaks from the java build path in the last two commits. I forgot to remove it from the java build path and look for invalid imports my bad. 

*If you are talking about DungeonTweaksCompat.class it's a class attatched directly to the battle towers commit*

The dungeon tweaks compat class is pure refelction currently.

jredfox:
the mod works without dungeon tweaks and compiles without dungoen tweaks I don't see a problem with the pr sorry If I confused you.

jredfox:
Is there still something wrong with this pull? I removed invalid imports reformated code you said was maul-formed and it's pure reflection thus it's a soft dependency. Sorry if your just to busy to take a look and merge. **it's just that people want to play battle towers and dungeon tweaks without a crash this pull was to fix both old and new dungeon tweaks.**

AtomicStryker:
Sorry, i was busy and reading scattered commits like this isn't easy. I looked over the squashed merge again and apart from a superfluous refactoring you did (adding Random as parameter when all you do is pull it from another parameter World beforehand...) it looks fine. I've updated the curse version with the new build.

I hope you tested the result to be working runtime, because i did not.

jredfox:
Well the random was intended to be used from whatever the gen had in mind not necessary world.rand. If you want I can take a look at the random and see if world randomness is accepted because I couldn't find a parameter elsewhere that wasn't the world.

jredfox:
are tower positions in chunk x and chunk z? If not I think you should fix it on your end I believe originally you might have used world.rand anyways.

jredfox:
Yeah well I thought you were going to attempt to use random as vanilla minecraft does but, I was wrong. You use world.rand which in the same pos will have different stuffs every time on both structures and spawners according to the logic of your mod. So just ignore it. I was able to convert it to vanilla minecraft at least the spawners but, unless the structure itself is vanilla mc style I wouldn't fix it.

vanilla mc logic is same seed same structure same loot same exact structure mobs etc....

AtomicStryker:
Yes, i am not making sure it generates deterministic because i don't care, but if that is a point for you i can understand the reasoning.

PR #242, opened 2018-12-15
[Ruins] resolve RuinsPositionsFile.txt concurrency issues

QuarterAnimal:
This change is intended to address a logged error reported in the Minecraft Forum by [kellixon](https://www.minecraftforum.net/forums/mapping-and-modding-java-edition/minecraft-mods/1282339-1-12-ruins-structure-spawning-system?comment=4502) ([twice](https://www.minecraftforum.net/forums/mapping-and-modding-java-edition/minecraft-mods/1282339-1-12-ruins-structure-spawning-system?comment=4528)) and [boyslittle](https://www.minecraftforum.net/forums/mapping-and-modding-java-edition/minecraft-mods/1282339-1-12-ruins-structure-spawning-system?comment=4579). Unfortunately, I'm not able to duplicate the error myself--its occurrence depends on unpredictable thread timing--so I can't confirm with certainty the reported problem is resolved. This change _does_ correct a couple real errors, though, and should go in regardless.

The primary issue, at least with regard to the error called out in the Forum, is the fact **RuinsGenerator.flushing** was incorrectly declared to be static, causing clashes between threads attempting to flush position data for different dimensions. Other somewhat-related fixes applied while I was in the neighborhood: checking and setting the **flushing** flag is now an atomic operation, contention between threads (load/flush) for access to **RuinsPositionsFile.txt** is eliminated, and set flags no longer leak after a flush. All these modifications enhance the thread-safety of RuinsGenerator.

With these changes in place, the code behaves as I believe was intended. Whether the intended behavior is appropriate, however, is another question. Specifically, I'm not sure flushing--or even ruin generation--should proceed at all until after loading is complete. I also don't think a flush in progress should inhibit subsequent flush attempts; seems to me the one in progress should be canceled in favor of the later one, in case more ruins were added. Performance might be improved by keeping track of what was added since the last flush instead of dumping everything every time. As this pull isn't meant to alter the existing design, though, these matters are best left for future consideration.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinGenerator.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinGenerator.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.Random;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import net.minecraft.block.Block;
 import net.minecraft.init.Blocks;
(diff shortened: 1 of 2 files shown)


AtomicStryker:
You have tested this? "synchronized (ruinsDataFile) { }" is actually completely new syntax to me. I googled and found out about synchronized blocks, neat. If i was going to rewrite this, i was probably going to use Fibers instead of Threads (aka CompletableFuture..). Eh, onto the wishing pile :)

QuarterAnimal:
I've tested it to the extent I know it's safe, and doesn't make things worse. As I said in the comments above, though, I have no way of knowing whether it eliminates the problem noted by kellixon and boyslittle, since I've never been able to reproduce it. I strongly suspect it will, but I won't know for sure until they're able to try it out.

PR #247, opened 2018-12-22
[Ruins] reset mod upon server launch

QuarterAnimal:
This change primarily fixes the error identified by boyslittle in [a recent Minecraft Forum post](https://www.minecraftforum.net/forums/mapping-and-modding-java-edition/minecraft-mods/1282339-1-12-ruins-structure-spawning-system?comment=4589). Ruins was initializing its worldgen database (**RuinsMod.generatorMap**) only once, on the mod load event. As a result, when running Singleplayer, Ruins continued to read from and write to the same world folder, even if the player switched to a different world. Notably, if that initial world was deleted and a new world created in the same game session, a "cannot find the path" error was thrown.

Ruins now listens for the **ServerAboutToStart** event and reinitializes its worldgen database accordingly, associating the selected world with its appropriate corresponding folder.

While I was in the area, I also made a correction to the code responsible for catching recursive generator calls. It was only looking at chunk coordinates and not considering the dimension in which the chunk was located. This could potentially (albeit rarely) lead to false positives; it's perfectly legitimate for the same chunk coordinates in different dimensions to be under generation simultaneously. Now, each dimension has its own "recursion watcher."

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinsMod.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinsMod.java
@@ -33,6 +33,7 @@ import net.minecraftforge.fml.common.IWorldGenerator;
 import net.minecraftforge.fml.common.Mod;
 import net.minecraftforge.fml.common.Mod.EventHandler;
 import net.minecraftforge.fml.common.event.FMLInitializationEvent;
+import net.minecraftforge.fml.common.event.FMLServerAboutToStartEvent;
 import net.minecraftforge.fml.common.event.FMLServerStartingEvent;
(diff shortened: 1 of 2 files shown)


AtomicStryker:
I did give you collaborator status, right? You should be able to push to the repo directly, if you so choose.

QuarterAnimal:
I could, but I wasn't sure that was your intent. Should I do that from now on? Maybe just for bug fixes?

AtomicStryker:
You want my intent?! You're going to need a ritual dagger and a couple of sacrificial animals to divine that.

Also, yes. So you can commit bugfixes. Githut has code reviews, right? You can always request one of those from me.

PR #248, opened 2018-12-23
[Ruins] allow fractional weights and chances

QuarterAnimal:
This changes the following elements so they no longer need be whole numbers:
- configured chances for biome-specific spawns--specific_forest=**98.5**
- template spawn weights--weight=**0.5**
- chances for template rule occurrence--rule3=0,**2.5**,dirt
- template rule block weights--rule5=0,100,cobblestone,**0.1***mossy_cobblestone

Whole numbers can still be used, of course, so existing templates aren't affected. Note the following have _not_ been changed and must remain integers:
- template rule variant weights--rule7=**3***0,100,gravel
- template rule variant group repeat counts--**5***rule9=0,100,sand

In addition, an existing bug causing specific spawn chances to be 1% lower than what was specified in **Ruins.txt** was fixed.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
@@ -22,10 +22,10 @@ import net.minecraft.world.biome.Biome;
 
 class FileHandler
 {
-    private final static int COUNT = 0, WEIGHT = 1, CHANCE = 2;
+    private final static int WEIGHT = 0, CHANCE = 1;
     private final HashMap<String, HashSet<RuinTemplate>> templates = new HashMap<>();
     private final int dimension;
-    private final HashMap<String, int[]> vars = new HashMap<>();
+    private final HashMap<String, double[]> vars = new HashMap<>();
 
     int triesPerChunkNormal = 6, triesPerChunkNether = 6;
     float chanceToSpawnNormal = 10, chanceToSpawnNether = 10;
@@ -139,11 +139,10 @@ class FileHandler
                 }
             }
 
-            // after all templates are loaded, calculate biome template counts and weights
+            // after all templates are loaded, calculate biome template weights
             for (String bname: templates.keySet())
             {
-                int[] val = new int[3];
-                val[COUNT] = templates.get(bname).size();
+                double[] val = new double[2];
                 vars.put(bname, val);
                 recalcBiomeWeight(bname);
             }
@@ -176,18 +175,15 @@ class FileHandler
     {
         try
         {
-            int rand = random.nextInt(vars.get(biome)[WEIGHT]);
-            int oldval = 0, increment = 0;
+            double rand = random.nextDouble()*vars.get(biome)[WEIGHT];
             RuinTemplate retval = null;
             for (RuinTemplate ruinTemplate : templates.get(biome))
             {
                 retval = ruinTemplate;
-                increment += retval.getWeight();
-                if ((oldval <= rand) && (rand < increment))
+                if ((rand -= retval.getWeight()) < 0)
                 {
-                    return retval;
+                    break;
                 }
-                oldval += retval.getWeight();
             }
             return retval;
         }
@@ -199,8 +195,8 @@ class FileHandler
 
     boolean useGeneric(Random random, String biome)
     {
-        int[] val = vars.get(biome);
-        return RuinsMod.BIOME_ANY.equals(biome) || (val != null && random.nextInt(100) + 1 >= val[CHANCE]);
+        double[] val = vars.get(biome);
+        return RuinsMod.BIOME_ANY.equals(biome) || (val != null && random.nextDouble() >= val[CHANCE]);
     }
 
     private void loadSpecificTemplates(PrintWriter pw, File dir, String bname) throws Exception
@@ -229,7 +225,7 @@ class FileHandler
     private void recalcBiomeWeight(String biomeName)
     {
         final Iterator<RuinTemplate> i = templates.get(biomeName).iterator();
-        int[] val = vars.get(biomeName);
+        double[] val = vars.get(biomeName);
         val[WEIGHT] = 0;
         while (i.hasNext())
         {
@@ -238,7 +234,7 @@ class FileHandler
         vars.put(biomeName, val);
     }
 
-    private static final Pattern patternSpecificBiome = Pattern.compile("\\s*specific_(\\w+)\\s*=\\s*(\\w+)\\s*(?:#|$)");
+    private static final Pattern patternSpecificBiome = Pattern.compile("\\s*+specific_([^\\s=]++)\\s*+=\\s*+([^\\s#]++)");
 
     private void readPerWorldOptions(File dir, PrintWriter ruinsLog) throws Exception
     {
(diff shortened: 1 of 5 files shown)


QuarterAnimal:
While this does have a minor bugfix in it, it's primarily a new feature (well...modifications to existing features). I want to run the idea by you before merging, in case you find fractions offensive.

Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
```diff
@@ -176,18 +175,15 @@ RuinTemplate getTemplate(Random random, String biome)
     {
         try
         {
-            int rand = random.nextInt(vars.get(biome)[WEIGHT]);
-            int oldval = 0, increment = 0;
+            double rand = random.nextDouble()*vars.get(biome)[WEIGHT];
```
AtomicStryker, line 178: formatting

Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
```diff
@@ -336,10 +332,10 @@ else if ((matcher = patternSpecificBiome.matcher(read)).lookingAt())
                     bgb = Biome.REGISTRY.getObject(rl);
                     if (bgb != null && bgb.getRegistryName().getResourcePath().equals(matcher.group(1)))
                     {
-                        int[] val = vars.get(bgb.getRegistryName().getResourcePath());
+                        double[] val = vars.get(bgb.getRegistryName().getResourcePath());
                         if (val != null)
                         {
-                            val[CHANCE] = Integer.parseInt(matcher.group(2));
+                            val[CHANCE] = Math.min(Math.max(Double.parseDouble(matcher.group(2))/100, 0), 1);
```
AtomicStryker, line 338: Is zero a "working" default?

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
```diff
@@ -906,7 +907,7 @@ else if (line.startsWith("biomeTypesToSpawnIn"))
                 else if (line.startsWith("weight"))
                 {
                     String[] check = line.split("=");
-                    weight = Integer.parseInt(check[1]);
+                    weight = Math.max(Double.parseDouble(check[1]), 0);
```
AtomicStryker, line 910: Does a weight of 0 do anything?

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
```diff
@@ -88,7 +88,7 @@ public RuinTemplateRule(PrintWriter dpw, RuinTemplate r, String rule, boolean de
             throw new Exception("No blockIDs specified for rule [" + rule + "] in template " + owner.getName());
         }
         condition = Integer.parseInt(blockRules[0]);
-        chance = Integer.parseInt(blockRules[1]);
+        chance = Math.min(Math.max(Double.parseDouble(blockRules[1])/100, 0), 1);
```
AtomicStryker, line 91: Again, 0 chance ... does that work?

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
```diff
@@ -754,7 +754,7 @@ private int getBlockNum(Random random)
     {
         // random selection using weights assigned in config file
         int blockIndex = 0;
-        for (int selector = random.nextInt(blockWeightsTotal); (selector -= blockWeights[blockIndex]) >= 0; ++blockIndex);
+        for (double selector = random.nextDouble()*blockWeightsTotal; (selector -= blockWeights[blockIndex]) >= 0; ++blockIndex);
```
AtomicStryker, line 757: formatting (i know the file isn't fully formatted with my template i know ...)

Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
QuarterAnimal: What's the issue?

Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
QuarterAnimal: Yes, as it always was (though the canned Ruins.txt config file sets them all to 75%). Zero is a valid value, effectively selecting only generic spawns for that biome. I'm not changing any policy here, thereby ensuring existing configurations behave as before without modification.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
QuarterAnimal: It means the template won't be selected for worldgen spawing. Whether there's a point to that or not, I can't say (for temporarily disabling a template, maybe?), but it's always been a valid value and still is. The only thing I'm doing now is guarding against negative weights, which used to get through and screw up the selection process.

I can't think of any particularly compelling reason to disallow 0--it works, and has the expected effect.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
QuarterAnimal: Zero works here, too. It's always been a valid value and has the expected effect--effectively rendering the rule a no-op (or preserveBlock, in Ruins terms). The only change is I'm clamping the value to be between 0% and 100%, inclusive (though values outside that range don't have the same deleterious effects as invalid weights).

QuarterAnimal: I did test these, by the way, if that's what you're asking.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
QuarterAnimal: Again, not sure what the issue is. I'm guessing it's got something to do with whitespace...do you always put spaces around binary operators?

Ruins/src/main/java/atomicstryker/ruins/common/FileHandler.java
AtomicStryker: Not really an issue, formatting would leave it as "random.nextDouble() * vars.get(biome)[WEIGHT];"

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
AtomicStryker: "nextDouble() * blockWeightsTotal" the formatting template would do this, yes. I am a fan of mandatory formatting by the way, because when someone eventually formats everything the diff is illegible. It's a problem at work too.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
AtomicStryker: Good good, i didn't want to risk a zero division somewhere.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
QuarterAnimal: Got it. I don't have any strong stylistic preferences, myself, so I'll adapt to the environment.

Issue #285, opened 2019-08-30
[1.12.2] [InfernalMobs 1.7.6] Mob gains 4 HP on every save load


scearcrovv:
### Description
Every time the save game is loaded, a mob gains 4 max HP.

The specific one is a Baby Zombie with now 61 HP and the following modifiers:
1UP, Berserk, Bulwark, Fiery, Regen, Rust, Sprint, Sticky, Strom, Vengeance, Webber

### Versions
minecraft@1.12.2, FML@8.0.99.99, forge@14.23.5.2836, mcp@9.42, infernalmobs@1.7.6, waila@1.8.26, jei@4.15.0.289

InfernalMobs was compiled from an up-to-date (d5b4d541) fork of the source.

AtomicStryker:
Likely a side effect of MC changes. The code for the "max health hack" stems from a time when max health was not persisted per-entity, but MC might do that now. So every time Infernal Mobs "hacks" a newly loaded or spawned entity, the max health extension happens again.

scearcrovv:
That's interesting. Is there anything we can do? I mean, it's a really small bug that I only noticed because I reloaded this particular savegame so many times.

AtomicStryker:
https://github.com/AtomicStryker/atomicstrykers-minecraft-mods/commit/2de37e82ed68d3cc0ba72b66ef145d65c38038ec

Its fixed in the current version atleast

PR #286, opened 2019-08-30
Implemented InfernalMobs Hwyla/WAILA integration.

scearcrovv:
The implementation only adds the modifiers to the body of the WAILA
tooltip window, name pre- and suffixes need more work.

Changes:

 - Added Hwyla Plugin and Registrar classes. Registrar adds two options:
   - infernalmobs.mods: Toggle the plugin on/off.
   - infernalmobs.linebreak: Inserts linebreaks in the modifier listing
     if a line would exceed 40 characters.

- Added WailaRegistrar.preInit call to InfernalMobsCore.preinit.

- Added necessary Hwyla dependency and repository to build.gradle.

### Note:
I did this for myself but thought you might be interested. I could also port this to 1.14.x if need be.

Also, I have no idea why GitHub shows **1895** changes in InfernalMobsCore.java, I literally _added three lines_ one of which is blank. IntelliJ showed me two changes during the commit.

### Screenshot
With and without linebreak. (This is incidentally the mob from Issue #285, at two different points)
HwylaPreview

### DIFF
--- a/InfernalMobs/build.gradle
+++ b/InfernalMobs/build.gradle
@@ -30,3 +30,16 @@ curseforge {
         }
     }
 }
+
+repositories {
+    maven {
+        name "Tehnut Maven FS"
+        url "https://tehnut.info/maven"
+    }
+}
+
+apply plugin: 'maven'
+
+dependencies {
+    deobfCompile "mcp.mobius.waila:Hwyla:1.8.26-B41_1.12.2"
+}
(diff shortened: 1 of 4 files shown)


AtomicStryker:
Does this need to be part of Infernal Mobs? I am *extremely* hesitant to add foreign party dependencies to the build file (had a bad experience with RL mod loader).

My question is, could this not be a build completely seperate from Infernal Mobs? I would quite happily expose a public API in the mod for the methods you need, instead of integrating something for which i cannot ascertain long term support in my "own" builds.

scearcrovv:
Of course, I understand your hesitation very well!

And yes, it is very much possible and I would even say significantly better to completely separate the plugin from IM with an appropriate API. This would truly enable long term support given the rather short live span of Minecraft mods while also enabling people to implement plugins for other interface mods such as The One Probe.

I would be very happy to make the plugin for Hwyla and publish it on Curse.

PR #297, opened 2019-10-11
Partially translated on Russian

NightKosh:


NightKosh:
Hello. Could you please provide an example of translation preffix with an explanation of which part of it corresponds to the translationparts? It is needed to me to be sure that all will be correctly translated.

AtomicStryker:
Sorry, i don't quite understand the question. If you are asking about how Prefixes and Suffixes work, the english translation file itself should be example aplenty.

Issue #300, opened 2019-10-18
[1.12.2][BattleTowers 1.6.5] BattletowerPositionsFile writes to wrong save folder


Charles445:
Create or load a save (World A)
Save and quit to title
Create or load a different save (World B)
World B's tower positions will now be writing to World A's folder instead of World B's folder.

If I understand correctly, this will mess with the spawning of the battle towers in some capacity.
I don't think servers will run into this bug, but single player definitely does.

AtomicStryker:
I will keep this in mind should i update the mod to 1.14, but i won't do any fixing for 1.12. The workaround is to not hotswap saves without restarting MC.

Thank you for reporting it.

xxlio109:
pls 1.12 FIX~~

PR #302, opened 2019-10-21
Fix cross save and cross dimensions bugs

Charles445:
Fix for #300 
This also fixes all dimensions reading and writing from the overworld positions file instead of their own.


AtomicStryker:
Slight formatting issues in the WordHandler file, ill fix them

AtomicStryker:
Thanks for the contribution!

PR #305, opened 2019-10-27
Fix separator, MP to SP edge case, curseforge request

Charles445:
Apparently pathSeparator is for separating 'PATH' values, that one caught me off guard.

Also a fix for an edge case where if you play multiplayer then singleplayer, the worldhandles still don't get cleared, causing writes to the root modpack folder. 

I tested this branch as best I could on single and multi, and it appears to be working fine.
(also I tried to get the formatting right, hope it worked)

Could you update the 1.12.2 BattleTowers on curseforge with this? I'm pretty sure the pack RLCraft is planning on updating soon, and BattleTowers has a large presence in it.


AtomicStryker:
Sure, pushed it to curse.

Issue #303, opened 2019-10-23
[Ruins] PreserveBlocks/Prefixes (!?) in 1.14.*


Tiscan76:
I'm kinda new to template-making and I'm currently trying to create some templates for a 1.14.4 modpack.

While I figured out how to use Loottables and how to create Spawners, there doesn't seems to be a way to preserve exsisting blocks?

From what I understand, there were 3 ways to force the mod to preserve/replace blocks in the 1.12 and prior.

By using "preserveBlock" in a rule, or by making blocks foreground/background blocks using the question-/exclamation mark as a prefix in the block-name.

I tried a few things but can't get this to work. I even decided to take a look at the Java-file but as I don't know much Java i'm not sure if what i found is correct. As far as I can tell, 1.12 setup an array called "preservePolicy" in RuinTemplateRule.java which handled what blocks should be preserved. And this seems to be gone completely.

Sorry if i'm just missing something obvious.

AtomicStryker:
To QuarterAnimal:

I have tested the "!rule" "foreground block" notation which according to your docs:

 "A "!" prefix does the opposite, creating a foreground
# block which ONLY replaces air and preserved blocks."

It does not appear to work as intended in 1.14 (maybe i broke it?). A block with ! prefix overwrites everything in the world, both preserved blocks and those than aren't.

QuarterAnimal:
Sorry...I'm just seeing this now. I'll get on it tonight.

QuarterAnimal:
To AtomicStryker:
Yes, support for foreground/background blocks was, among other things (such as block weighting), removed as part of the 1.14 migration. I can add it back in, but I'm not sure implementing it as a !/? prefix is appropriate now--it seems you went the way of specifying everything within an NBT compound. Maybe you'd rather see a Ruins-specific tag within the block NBT for such stuff? That'd be more consistent with what you've already got and (arguably) more extensible. Let me know.

AtomicStryker:
You can check RuleStringNbtHelper#StringFromBlockState - doing everything in a NBT compound is actually fully vanilla, and i intend to stick to it for compatibility reasons if nothing else.

As you can also see i've embedded the tile entity nbt as a tag inside the block nbt - so by all means feel free to embed your implementation metadata too.

Tiscan76:
Thanks for the work you put into it.

Here is a little screenshot that shows why it is important to be able to keep blocks intact:

[...]

PR #307, opened 2019-11-03
[Ruins] reinstate old preserveBlock functionality, et al. (issue #303)

QuarterAnimal:
These changes introduce a new pseudo-block--**ruins:null**--for use in rules. It functions as the old **preserveBlock** did, leaving the existing block at its location intact. The default rule0 is now defined as **ruins:null** instead of **minecraft:air**. Example, this rule3 is 50% cobble, 50% nuthin':
`rule3={Name:"cobblestone"}{Name:"ruins:null"}`

I also added a new tag--"Ruins"--to hold all Ruins-specific block parameters, as well as implementations for three such parameters: "weight", "bonemeal", and "entity". Example, this rule5 is 70% oak sapling, 10% fertilized birch sapling (many of which are grown into trees), and 20% spruce sign:
`rule5={Name:"oak_sapling",Ruins:{weight:3.5}}{Name:"birch_sapling",Ruins:{weight:.5,bonemeal:1}}{Name:"spruce_sign",Ruins:{entity:{Text2:"\"Ceci n'est pas\"",Text3:"\"un arbre.\""}}}`

All, some, or none of these parameters may be specified for a particular block. "weight" functions as it did in 1.12, defaulting to 1; it doesn't have to be an integer, but it can't be negative. "bonemeal" functions as the old "-addbonemeal" suffix, except you can now smack that sucker with multiple bonemeals by specifying a number greater than 1; it can't be negative, and defaults to 0. "entity" is a replacement for "ruinsTE", bringing all the Ruins parameters under one easily-extensible umbrella. However, "ruinsTE" is still supported as before, though it's deprecated.

Also: NBT parsing was fixed to properly handle braces within quoted strings, and block entity specifications no longer require an "id" tag (as seen in example rule5, above).

What I _didn't_ do was implement the foreground/background thing. I'll wait to see if there's any demand for it.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplate.java
@@ -1,7 +1,23 @@
 package atomicstryker.ruins.common;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Random;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.block.Blocks;
+import net.minecraft.block.IGrowable;
+import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 import net.minecraft.world.biome.Biome;
@@ -10,13 +26,6 @@ import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.fml.ModList;
 import net.minecraftforge.fml.common.registry.GameRegistry;
 import net.minecraftforge.registries.IForgeRegistry;
-
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileReader;
-import java.util.*;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 public class RuinTemplate {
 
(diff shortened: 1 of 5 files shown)


QuarterAnimal:
I'm not going to merge this myself until the "ruinsTE" disposition is vetted.

AtomicStryker:
Great work! My only concern is that "entity" is too generic a tag and we suffer some kind of aliasing issue in the future. What if a mod blockstate or TE uses that tag key?

QuarterAnimal:
Context. I put all the parameter tags under the "Ruins" TAG_Compound specifically to avoid any such conflicts now and in the future...if an "entity" tag appears in some other context, it's fine.

I'm not entirely thrilled about the name "entity," either, but more on aesthetic rather than technical grounds. I originally called it "block_entity" (which is what used to be tile entities are called now), but I thought that overly verbose and a tad redundant. I also considered "nbt," but that's how it's formatted, not what it is. "data," maybe? Eh. I'm open to suggestions--it's almost trivial to change.

AtomicStryker:
I would suggest something that links the tag to this mod. So if someone has a template rule string with no other context, some googling would deliver him somewhere useful. "entity" does not fulfill that ideal - maybe "ruins-data"?

QuarterAnimal:
As it is, a block specification looks like this:
`{Name:"...",Properties:{...},Ruins:{weight:1,bonemeal:0,entity:{...}}}`
So if someone has a template rule string, she necessarily has the context--the "entity" tag only lives inside the "Ruins" tag. Expecting to Google "entity" by itself for information would be like expecting to Google "x" by itself for information on that tag. But...if you're intent on stamping each individual member with its context, it's an easy enough change, and I don't feel all that strongly about it. However, if googleability is the motivation, it should probably be applied to all the parameters. How about "ruins_weight," "ruins_grow," and "ruins_block_entity" (because that's exactly what they are)?

AtomicStryker:
Ooh i was not aware of that. My bad then. Feel free to merge

Issue #318, opened 2020-01-02
[1.14.4] Log spam


Gbergz:
Logs and debug log is being spammed by this, i have no idea why but it does. Any ideas?

Logs Spam: [https://gist.github.com/Gbergz/bb10d5a696959a835ec963faa88d7d48](https://gist.github.com/Gbergz/bb10d5a696959a835ec963faa88d7d48)

**MC-Version** 1.14.4
**Forge Version** 28.1.107
**Mod-Version:** ruins-1.14.4.4.jar

QuarterAnimal:
I'll add a throttle on the deprecation warnings. Currently, the warning is emitted every time the "old" block entity specification (**ruinsTE**) is encountered in the template files. I suppose there's no point in seeing the message more than once, though, aside from letting you know how many lines should be edited. Note it's just a deprecation warning, so you can ignore it...for now.

There are some legit errors flagged in that log, however. At least 2 invalid rules appear in your set of templates. Multiple block entities for a single block? That's not a thing.

PR #320, opened 2020-01-06
[Ruins] address user feedback (#318)

QuarterAnimal:
In issue #318 , Gbergz describes an overabundance of **ruinsTE** deprecation warnings in the log file. The number of warnings is now capped at 4; subsequent notifications are sent to the debug log only. I think it's appropriate to still flag _every_ occurrence in the debug log, as that's what the debug log is for, but the general log is more useful without the clutter.

In [Minecraft Forum](https://www.minecraftforum.net/forums/mapping-and-modding-java-edition/minecraft-mods/1282339-1-15-1-ruins-structure-spawning-system?comment=4733), argonz mentions ruins spawning in trees. While there are workarounds, the intended functionality which would avoid the problem is broken. During the 1.13.2 migration, Block.isFoliage() was introduced as a way to check for ignorable plant blocks. This is a Forge extension which--at least in 1.15.1--isn't implemented for any vanilla blocks. As a result, only air, snow, and cobwebs were ignored. Now the check is based on block materials and properly recognizes plants and leaves.

### DIFF
--- a/Ruins/build.gradle
+++ b/Ruins/build.gradle
@@ -17,7 +17,7 @@ apply plugin: 'eclipse'
 apply plugin: 'maven-publish'
 
 def mcversion = '1.15.1'
-version = '1.15.1.1'
+version = '1.15.1.2'
 group = 'atomicstryker.ruins'
 archivesBaseName = 'ruins'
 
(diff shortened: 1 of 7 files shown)


AtomicStryker:
Good catch, i didn't check the implementation of the foliage check. I swear it used to be Material.plant at some point.

Another thing: Wouldn't it be a good idea to fix "ruinsTE" occurences in the default templateset? That's what most people will see, after all

QuarterAnimal:
Yes, that would be a smashing idea. I considered doing that when I submitted the pull to deprecate **ruinsTE** in the first place, but decided to wait and see if you approved of the change beforehand...and then forgot to go back and update the templates. Here I am using those templates to test my code--you'd think that'd jog my memory. I'm getting old. Anyway, I'm on it.

PR #353, opened 2020-05-23
[Ruins]Fix an issue that prevent loot tables from mods being used.

dong031001:
```
String[] s = dataString.split(":");
int targetCount = s.length > 1 ? Integer.valueOf(s[2].split("-")[0]) : 0;
```
This block of codes produces
```
[19:52:38] [Server thread/INFO] [STDERR]: [atomicstryker.ruins.common.RuinTemplate:doBuild:336]: java.lang.NumberFormatException: For input string: "chests/chestloot1"
[19:52:38] [Server thread/INFO] [STDERR]: [atomicstryker.ruins.common.RuinTemplate:doBuild:336]: 	at java.lang.NumberFormatException.forInputString(Unknown Source)
[19:52:38] [Server thread/INFO] [STDERR]: [atomicstryker.ruins.common.RuinTemplate:doBuild:336]: 	at java.lang.Integer.parseInt(Unknown Source)
[19:52:38] [Server thread/INFO] [STDERR]: [atomicstryker.ruins.common.RuinTemplate:doBuild:336]: 	at java.lang.Integer.valueOf(Unknown Source)
[19:52:38] [Server thread/INFO] [STDERR]: [atomicstryker.ruins.common.RuinTemplate:doBuild:336]: 	at atomicstryker.ruins.common.RuinTemplateRule.doSpecialBlock(RuinTemplateRule.java:609)
```
when `ChestGenHook:neuroflow:chests/chestloot1:5-3` is parsed.

### DIFF
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
@@ -606,10 +606,7 @@ public class RuinTemplateRule
         else if (dataString.startsWith("ChestGenHook:"))
         {
             String[] s = dataString.split(":");
-            if(s.length > 3){
-                s[1] = String.join(":", s[1], s[2]);
-                s[2] = s[3];
-            }
+            if(s.length > 3) s[1] = String.join(":", s[1], s[2]);
             int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-1].split("-")[0]) : 0;
             addChestGenChest(world, random, x, y, z, s[1], targetCount, blocknum, rotate);
         }
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
@@ -606,7 +606,7 @@ public class RuinTemplateRule
         else if (dataString.startsWith("ChestGenHook:"))
         {
             String[] s = dataString.split(":");
-            int targetCount = s.length > 1 ? Integer.valueOf(s[s.length-2].split("-")[0]) : 0;
+            int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-2].split("-")[0]) : 0;
             addChestGenChest(world, random, x, y, z, s[1], targetCount, blocknum, rotate);
         }
         else if (dataString.startsWith("IInventory;"))
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
@@ -606,7 +606,7 @@ public class RuinTemplateRule
         else if (dataString.startsWith("ChestGenHook:"))
         {
             String[] s = dataString.split(":");
-            int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-2].split("-")[0]) : 0;
+            int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-1].split("-")[0]) : 0;
             addChestGenChest(world, random, x, y, z, s[1], targetCount, blocknum, rotate);
         }
         else if (dataString.startsWith("IInventory;"))
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
@@ -607,7 +607,8 @@ public class RuinTemplateRule
         {
             String[] s = dataString.split(":");
             int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-1].split("-")[0]) : 0;
-            addChestGenChest(world, random, x, y, z, s[1], targetCount, blocknum, rotate);
+            String lootTable = s.length > 3 ? s[1]+s[2] : s[1];
+            addChestGenChest(world, random, x, y, z, lootTable, targetCount, blocknum, rotate);
         }
         else if (dataString.startsWith("IInventory;"))
         {
--- a/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
+++ b/Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
@@ -606,9 +606,12 @@ public class RuinTemplateRule
         else if (dataString.startsWith("ChestGenHook:"))
         {
             String[] s = dataString.split(":");
+            if(s.length > 3){
+                s[1] = String.join(":", s[1], s[2]);
+                s[2] = s[3];
+            }
             int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-1].split("-")[0]) : 0;
-            String lootTable = s.length > 3 ? s[1]+s[2] : s[1];
-            addChestGenChest(world, random, x, y, z, lootTable, targetCount, blocknum, rotate);
+            addChestGenChest(world, random, x, y, z, s[1], targetCount, blocknum, rotate);
         }
         else if (dataString.startsWith("IInventory;"))
         {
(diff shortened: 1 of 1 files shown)


Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
```diff
@@ -606,7 +606,7 @@ else if (dataString.startsWith("HardChest"))
         else if (dataString.startsWith("ChestGenHook:"))
         {
             String[] s = dataString.split(":");
-            int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-2].split("-")[0]) : 0;
+            int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-1].split("-")[0]) : 0;
             addChestGenChest(world, random, x, y, z, s[1], targetCount, blocknum, rotate);
```
QuarterAnimal, line 610: You're only passing the namespace portion of the loot table resource location, s[1], to addChestGenChest(). So, for example, "foo:chests/bar" will be interpreted as "minecraft:foo"...not what you want.

A simpler solution might be to leave the original code as-is, but just massage the s array a bit before using it, glomming the pieces of the loot table location together:

```suggestion
            if (s.length > 3)
            {
                s[1] = String.join(":", s[1], s[2]);
                s[2] = s[3];
            }
```

QuarterAnimal, line 611: You're only passing the namespace part of the loot table location, s[1], to addChestGenChest(). So, for example, a location of "neuroflow:chests/chestloot1" will be interpreted as "minecraft:neuroflow"...not what you want.

Maybe it'd be simpler to leave the existing code as-is and just massage the s array a bit, glomming the parts of the loot table location together, before using it? Something like:
```
String[] s = dataString.split(":");
if (s.length > 3)
{
    s[1] = String.join(":", s[1], s[2]);
    s[2] = s[3];
}
```

dong031001: Right,right, that will work.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
```diff
@@ -607,7 +607,8 @@ else if (dataString.startsWith("ChestGenHook:"))
         {
             String[] s = dataString.split(":");
             int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-1].split("-")[0]) : 0;
-            addChestGenChest(world, random, x, y, z, s[1], targetCount, blocknum, rotate);
+            String lootTable = s.length > 3 ? s[1]+s[2] : s[1];
```
QuarterAnimal: You need to restore the colon between the namespace and path here, so s[1] + s[2] should be s[1] + ":" + s[2].

QuarterAnimal: Never mind...it looks like I wasn't looking your the latest code.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
```diff
@@ -606,9 +606,12 @@ else if (dataString.startsWith("HardChest"))
         else if (dataString.startsWith("ChestGenHook:"))
         {
             String[] s = dataString.split(":");
+            if(s.length > 3){
+                s[1] = String.join(":", s[1], s[2]);
+                s[2] = s[3];
+            }
             int targetCount = s.length > 2 ? Integer.valueOf(s[s.length-1].split("-")[0]) : 0;
```
QuarterAnimal, line 610: If you're going to keep your change to use s[s.length-1] here instead of the original code's s[2], you can get rid of line 611. I added that line under the assumption nothing else was changing. Either way works, but you don't need both.

QuarterAnimal:
One overall issue I have--I'm not sure it's within the scope of this pull, though, but from a user perspective it's bound to cause confusion. The **IInventory** directive also supports a specification of "ChestGenHook"; however, despite having the same name, it's parsed and processed differently from the **ChestGenHook** directive. IInventory's ChestGenHook can't support mod loot tables or draw counts other than 1 (despite comments to the contrary).

Should IInventory's code be updated, too? You could take the position it's already kind of broken, so fixing it isn't your immediate responsibility. Without updating IInventory, though, you'll only be able to use mod loot tables to fill vanilla chests--you can't fill, say, barrels, furnaces, or mod containers.

Something to think about.

Ruins/src/main/java/atomicstryker/ruins/common/RuinTemplateRule.java
dong031001: Got it. Was treating the array as LinkedList lmao.

dong031001:
> One overall issue I have--I'm not sure it's within the scope of this pull, though, but from a user perspective it's bound to cause confusion. The **IInventory** directive also supports a specification of "ChestGenHook"; however, despite having the same name, it's parsed and processed differently from the **ChestGenHook** directive. IInventory's ChestGenHook can't support mod loot tables or draw counts other than 1 (despite comments to the contrary).
> 
> Should IInventory's code be updated, too? You could take the position it's already kind of broken, so fixing it isn't your immediate responsibility. Without updating IInventory, though, you'll only be able to use mod loot tables to fill vanilla chests--you can't fill, say, barrels, furnaces, or mod containers.
> 
> Something to think about.

For me, as I will only use vanilla chests for looting, this problem isn't the one I would think about. However, It is still worthy to work on in possible restructuring of this mod for other users' benifits.
Just a possible way: can we pull an ItemStack[] using the vanilla-way of pulling items from a loot table, then distribute the itemstacks in the array to the container? I didn't look up the current code but I'd assume that's likely doable.

QuarterAnimal (commented):
I'm cool with this now, from a functional point of view.

Hopefully, AtomicStryker will chime in with some comments, but I can anticipate two of them. First, he'll want to see the established code format maintained--if you use Eclipse, see eclipse_formatter_export.xml in the repository's root folder. Second, he'll ask if you built and tested the change. If you're unable to do one or both of these, let me know and I'll help.

I'd also like to hear whether he considers this change complete without updating IInventory's ChestGenHook code. It wouldn't be difficult to add support for mod loot tables there, too--maybe a one-line change. Implementing the draw count, though, is more problematic, and I'd recommend against it (at least in this pull).

dong031001:
> I'm cool with this now, from a functional point of view.
> 
> Hopefully, AtomicStryker will chime in with some comments, but I can anticipate two of them. First, he'll want to see the established code format maintained--if you use Eclipse, see eclipse_formatter_export.xml in the repository's root folder. Second, he'll ask if you built and tested the change. If you're unable to do one or both of these, let me know and I'll help.
> 
> I'd also like to hear whether he considers this change complete without updating IInventory's ChestGenHook code. It wouldn't be difficult to add support for mod loot tables there, too--maybe a one-line change. Implementing the draw count, though, is more problematic, and I'd recommend against it (at least in this pull).

Built and worked. For the code style one, I am not using Eclipse, but I think the line I added won't be a big issue.

AtomicStryker:
Hi, sorry but gmail put all github mails (and other importants mails ... T_T) into my junk folder for the last month and i've only stumbled on this today.

I'd defer to QuarterAnimal on this. Yes please consistent formatting if at all useful.

dong031001:
Is there any attempt to merge this pull? I am actually quite care about this Pull as a major update of my modpack, indeed, is dependent on this function and, although there are ways to work around it, a line change or two will make it a lot easier.

tyon2006:
Hey, I see this issue was merged into the code in 2020, but the latest posted version on Curseforge is from 2019. Forgive my lack of understanding, but is this issue in the Curseforge current 1.12.2 or not?

AtomicStryker:
It was not part of any release. I have pushed a build to curse for you which includes the merged pull request.

Issue #350, opened 2020-05-05
Ruins do not spawn in Terraforged world type


marcusbeckerman:
Using the Terraforged mod and generating a Terraforged world type, ruins do not spawn naturally. It works just fine in a default world, the nether, and the end but not in Terraforged.

QuarterAnimal:
Ruins hooks into default vanilla chunk decoration--namely, **ChunkGenerator.decorate()**--while TerraForged overrides it entirely. It is safe to use both mods together, but Ruins structures cannot appear in TerraForged dimensions; e.g., the overworld.

This is always a potential issue whenever using multiple custom chunk generation mods in general, not just these two, and there's no practical way to resolve it. Ruins could sink its coremod hook into a different location in the code, but that just moves the issue rather than solves it and is bound to run afoul of other mods in short order. It'd certainly be possible to create a compatibility module or bridge mod specific to Ruins and TerraForged, but that sort of thing is fragile, difficult to maintain, and probably of limited utility.

If Forge were to add a chunk decoration event, it might help in this case, though I suspect that's not going to happen. An established generic scheme for custom decoration already exists, obviating the need for such an event. For better or worse, Ruins doesn't use it. Maybe it should, but that would be a radical redesign, essentially making it an entirely different mod, unable to support many of its current features.

For the time being at least, I think we just have to live with the incompatibility, unless AtomicStryker is considering a complete overhaul of Ruins in the near future.

QuarterAnimal:
As of version 1.15.2.1, Ruins works with TerraForged.

Sanysvo:
Its not work( i am download last ver. of Ruins and Terra and not have any structure in my terra world

QuarterAnimal:
I just downloaded the latest versions of Forge, Ruins, and TerraForged, started up a fresh new world without making any config changes from the defaults, and I'm seeing Ruins on a TerraForged landscape. I did the same for a dedicated server installation and got the same results. Clearly the two mods _do_ work together.

What other mods are you using?

PR #354, opened 2020-05-26
[Ruins] update to Minecraft 1.15.2/Forge 31.2.0

QuarterAnimal:
These changes bring Ruins up to 1.15.2, eliminate the need for coremod injection, clean up a few compiler warnings, and put zombies on pirate ships.

This resolves issue #350, as Ruins can now spawn in TerraForged dimensions. It also addresses a couple issues raised in CurseForge comments, such as why ruins don't spawn with TerraForged (again with the TerraForged!), why drowned spawners don't work on pirate ships, and can i haz 1.15.2.


AtomicStryker:
Hi, same as in the other PR i did not see this at all, sorry. You removed the coremod dependency in the code, but you didn't remove the coremod files themselves?

QuarterAnimal:
Which files do you mean? The files **coremods.json** and **ruinstransformer.js** are deleted under [4172a06](https://github.com/AtomicStryker/atomicstrykers-minecraft-mods/pull/354/commits/4172a0659030061a811d8a06690d1d8e371cb511).

QuarterAnimal:
I didn't merge this one myself because I figure migration to a new version is significant enough to warrant waiting for your consent. Plus I'm not able to push to Curse, so you'd have to intervene anyway.

AtomicStryker:
Indeed they are, either i didn't scroll or i just missed them, my bad.

Technically 1.15.1 is is fully compatible to 1.15.2 (the reverse is probably also true), which is the reason i did not push new versions for 1.15.2

Let me merge this right now (i assume you tested this and made sure the new decoration hook works?), test it for 2 minutes, then ill push a version

AtomicStryker:
Playtested it, seems to work fine (but boy did MC annoy me with ocean-only worlds today). I will push this version to curse momentarily :)

AtomicStryker:
Done. From what i can tell of the download numbers the majority of players seems to be still using 1.12 ... by a HUGE margin

QuarterAnimal:
I can believe it. A lot of popular mods didn't make the 1.13 leap (yet?), so I expect 1.12 is going to be lingering for a while.

My primary motivation here was to make Ruins compatible with emerging custom chunk generator mods, like TerraForged. Since it's no longer a coremod, it should play nicely with those. I also hope that, by following the established vanilla chunk decoration scheme (kind of...), we're a lot less vulnerable to being broken in future Minecraft and Forge versions. I have a feeling another big wave of worldgen changes is coming.

PR #425, opened 2022-04-04
Enable light-level feature and support item tags

davenonymous:
Hey there!

This pull requests adds two features to dynamic lights:
- Varying light levels for items, e.g. redstone torches can emit less light instead of the hardcoded 15
- Support for configuring items that emit light using various item tags

## Varying light levels
The first two commits handle this feature.
The first commit gives the 3 BlockLit* an additional `power` blockstate property. (With this a few more "model missing" lines are being thrown at game start, since there is no blockstate definition present)
The second commit makes use of this property by allowing it to be configured in the existing config files.

## Item Tags
The third commit makes the two currently existing light sources aware of various `dynamiclights:*` item tags.
A pack or mod dev can now add various items by assigning tags to them. This is particularly useful when exact nbt matching isn't desired (e.g. if a torch gets renamed in an anvil it should still emit light). For exact NBT matches the original config system should be used.

The tags are:
```
dynamiclights:self (light level 15)
dynamiclights:self_1 (light level 1)
...
dynamiclights_self_8 (light level 8)
...
dynamiclights:self_15 (light level 15)
```
And the same with `dropped` instead of `self`.

The fourth commit adds a rather straight forward `dynamiclights:not_waterproof` tag.

This PR also includes some data tags that you might want to tweak or delete for better default values?

Hope these are useful improvements.
Happy to change and improve things in this PR further.

### DIFF
--- a/DynamicLights/src/main/java/atomicstryker/dynamiclights/server/DynamicLightSourceContainer.java
+++ b/DynamicLights/src/main/java/atomicstryker/dynamiclights/server/DynamicLightSourceContainer.java
@@ -6,6 +6,7 @@ import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.LiquidBlock;
 import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.block.state.properties.BlockStateProperties;
 
 import java.util.Map;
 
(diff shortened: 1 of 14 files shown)


AtomicStryker:
Super cool! I think i will definitely merge this. One request: Personally, i prefer explicit typing to "var" because one can tell at a glance (or in a web review tool) what type each variable is, rather than needing to know or an IDE to hint it. Could you exchange the vars for explicit types? For me? :)

davenonymous:
Cool!
Of course. I replaced the `var`s with the appropriate types.

Edit: Oh, I should also note that the item tags introduce a breaking change in the 1.18.2 version. They are handled slightly different there.
I've included the two lines that should be used in 1.18.2 instead.

AtomicStryker:
Just FYI, i touched it up very slightly (and used the vanilla torch light value of 14)

PR #487, opened 2023-11-11
fix logspam caused by missing models

Tfarcenim:
This fixes the several hundred lines of logspam that is caused by the lit blocks not having models, this PR gives them models and datagen so that doesn't happen.

Attached is a snippet of what I see when using the mod as a dependency, I'm fixing it on my end for now, but I made this PR so others could have the fix as well.

idea64_PKHUsPCT6l

### DIFF
--- a/DynamicLights/src/generated/resources/.cache/af6b71d80a7bad2e79730e7a7b72cf91265d7018
+++ b/DynamicLights/src/generated/resources/.cache/af6b71d80a7bad2e79730e7a7b72cf91265d7018
@@ -1,0 +1,4 @@
+// 1.20.1	2023-11-11T03:45:54.0868385	Block States: dynamiclights
+3dff9e1ec59415779195d2abfa9f07bc17b428a1 assets/dynamiclights/blockstates/lit_air.json
+3dff9e1ec59415779195d2abfa9f07bc17b428a1 assets/dynamiclights/blockstates/lit_cave_air.json
+e5a33b4a2db42195222fa1709b3d9533ac41e453 assets/dynamiclights/blockstates/lit_water.json
--- a/DynamicLights/src/main/java/atomicstryker/dynamiclights/server/datagen/ModBlockStateProvider.java
+++ b/DynamicLights/src/main/java/atomicstryker/dynamiclights/server/datagen/ModBlockStateProvider.java
@@ -1,0 +1,21 @@
+package atomicstryker.dynamiclights.server.datagen;
+
+import atomicstryker.dynamiclights.server.DynamicLights;
+import net.minecraft.data.PackOutput;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.client.model.generators.BlockStateProvider;
+import net.minecraftforge.common.data.ExistingFileHelper;
+
+public class ModBlockStateProvider extends BlockStateProvider {
+    public ModBlockStateProvider(PackOutput output, ExistingFileHelper exFileHelper) {
+        super(output, DynamicLights.MOD_ID, exFileHelper);
+    }
+
+    @Override
+    protected void registerStatesAndModels() {
+        final ResourceLocation airModel = new ResourceLocation("block/air");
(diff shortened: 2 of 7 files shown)


DynamicLights/src/generated/resources/.cache/af6b71d80a7bad2e79730e7a7b72cf91265d7018
```diff
@@ -0,0 +1,4 @@
+// 1.20.1	2023-11-11T03:45:54.0868385	Block States: dynamiclights
+3dff9e1ec59415779195d2abfa9f07bc17b428a1 assets/dynamiclights/blockstates/lit_air.json
```
AtomicStryker, line 2: I have zero experience with these, should src/generated/ files be commited? It feels like ModDatagen would create the files

DynamicLights/src/main/java/atomicstryker/dynamiclights/server/datagen/ModBlockStateProvider.java
```diff
@@ -0,0 +1,21 @@
+package atomicstryker.dynamiclights.server.datagen;
+
+import atomicstryker.dynamiclights.server.DynamicLights;
+import net.minecraft.data.PackOutput;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.client.model.generators.BlockStateProvider;
+import net.minecraftforge.common.data.ExistingFileHelper;
+
+public class ModBlockStateProvider extends BlockStateProvider {
```
AtomicStryker, line 14: comment would be nice as to what the purpose of this class is

DynamicLights/src/main/java/atomicstryker/dynamiclights/server/datagen/ModDatagen.java
```diff
@@ -0,0 +1,17 @@
+package atomicstryker.dynamiclights.server.datagen;
+
+import net.minecraft.data.DataGenerator;
+import net.minecraft.data.PackOutput;
+import net.minecraftforge.common.data.ExistingFileHelper;
+import net.minecraftforge.data.event.GatherDataEvent;
+
+public class ModDatagen {
```
AtomicStryker, line 13: class comment would be nice

DynamicLights/src/main/java/atomicstryker/dynamiclights/server/datagen/ModDatagen.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public static void start(GatherDataEvent event) {
+        DataGenerator dataGenerator = event.getGenerator();
+        PackOutput packOutput = dataGenerator.getPackOutput();
+        ExistingFileHelper existingFileHelper = event.getExistingFileHelper();
+        boolean client = event.includeClient();
+        dataGenerator.addProvider(client,new ModBlockStateProvider(packOutput,existingFileHelper));
```
AtomicStryker: spaces, formatting?

AtomicStryker (commented):
I am super stoked whenever i get a PR (in over a decade i have received  so few...).
Thank you for this!

One legit question and some formatting wah wah, if you would be so nice. Also feel free to add a thanks to yourself in the changelog :)

DynamicLights/src/generated/resources/.cache/af6b71d80a7bad2e79730e7a7b72cf91265d7018
Tfarcenim: Depends, some prefer to commit the files, some prefer to .gitignore and generate them locally.  I usually leave them in the repo so I can point people to what can be configured with a datapack.

DynamicLights/src/main/java/atomicstryker/dynamiclights/server/datagen/ModBlockStateProvider.java
Tfarcenim: Comment added.

PR #489, opened 2023-11-22
[BattleTowers] Added lang key for Golem defeated message

KameiB:
- Also added es_mx.lang file


AtomicStryker:
Looking good, thanks!

PR #531, opened 2025-01-29
Add Champions Unofficial compat

Hypocisy:
## Add champions Compat
- Add Champions Compat, if entity aready champion mob, stop process spawning entity to inf mob
- We should check if entity is champion, not only check champion entity's persistent data, because when champion created, entity does not have persistent data yet.
- in this case, I made champion spawn event to HIGHEST EventPriority，so forge should first let entity become champion, and then the inf mod will check if champion entity persistet has PersistentData, but no persistentData to check.
- Please see those pictures
- image
- image
- image


AtomicStryker:
First, thank you for participating

The "code cleaning" has made the PR much bigger and harder to read, sadly.
It would have maybe been better to split this into 2 PR, one for "functionality" and one for "code style changes"

Generally speaking, i am very hesitant to add hard dependencies to my mods because in the past, dependencies always stopped updating or updated later etc

Hypocisy:
it's not hard dependencies, it's soft at all.
with out champions, your mod still can run property

Hypocisy:
sorry for that, i will split it to 2 PR

Hypocisy:
@AtomicStryker Well, right now this pr only contain champion compat related code

InfernalMobs/src/main/java/atomicstryker/infernalmobs/common/InfernalMobsCore.java
```diff
@@ -114,6 +80,7 @@ public InfernalMobsCore() {
         classesForcedMap = new HashMap<>();
         classesHealthMap = new HashMap<>();
         modifiedPlayerTimes = new HashMap<>();
+        championLoaded = ModList.get().isLoaded("champions");
```
AtomicStryker: you can probably pull the modID from the actual ChampionsCore class, rather than hardcoding a magic string?

InfernalMobs/src/main/java/atomicstryker/infernalmobs/common/EntityEventHandler.java
```diff
@@ -31,16 +32,28 @@ public EntityEventHandler() {
...
-            if (!savedMods.isEmpty() && !savedMods.equals(InfernalMobsCore.instance().getNBTMarkerForNonInfernalEntities())) {
-                InfernalMobsCore.instance().addEntityModifiersByString((LivingEntity) event.getEntity(), savedMods);
-            } else {
-                InfernalMobsCore.instance().processEntitySpawn((LivingEntity) event.getEntity());
+        var entity = event.getEntity();
+        var level = event.getLevel();
+        // check level is server level first
+        if (!level.isClientSide()) {
+            // champions mob compat start
+            if (InfernalMobsCore.instance().isChampionLoaded() && ChampionsCompat.isChampionEntity(entity)) {
```
AtomicStryker: should add a boolean configurable that ALLOWS a mob to be both champion and infernal. masochists will ask for it

InfernalMobs/build.gradle
```diff
@@ -74,8 +94,33 @@ minecraft {
...
+            jarJar.ranged(it, "[${mixin_extras_version},)")
+        }
+    } else {
+        implementation("io.github.llamalad7.mixinextras:mixinextras-forge:${mixin_extras_version}")
+    }
+
+    annotationProcessor "org.spongepowered:mixin:${mixin_version}:processor"
+
+    compileOnly fg.deobf("curse.maven:champions-unofficial-1074990:${cf_champions}")
+    compileOnly fg.deobf("curse.maven:jade-324717:${cf_jade}")
```
AtomicStryker: whats jade? some kind of HUD? why depend on it? you dont appear to be importing any classes from it anywhere

InfernalMobs/gradle.properties
```diff
@@ -58,4 +58,14 @@ mod_authors=AtomicStryker
 # The description of the mod. This is a simple multiline text string that is used for display purposes in the mod list.
 mod_description=Randomly empower Mobs with Diablo-Style \nmodifiers and grant loot rewards for defeating them.
 
-curseforge_id=227875
\ No newline at end of file
+curseforge_id=227875
+
+# mod depends
+cf_champions=6133394
```
AtomicStryker: this is a specific version of champions, yes? please write how to get this number from curseforge files so this is maintainable

InfernalMobs/build.gradle
```diff
@@ -74,8 +94,33 @@ minecraft {
...
+        url = "https://www.cursemaven.com"
+    }
+    mavenLocal()
+    mavenCentral()
+}
+
 dependencies {
     minecraft "net.minecraftforge:forge:${minecraft_version}-${forge_version}"
+
+    compileOnly(annotationProcessor("io.github.llamalad7:mixinextras-common:${mixin_extras_version}"))
```
AtomicStryker: is all of this really needed? we are not compiling champions, we just want its jar and read one class from it, but i really dont have much experience with sponge/mixins

AtomicStryker (changes requested):
already much cleaner looking! i think there is still more in your PR than is required

InfernalMobs/src/main/java/atomicstryker/infernalmobs/common/InfernalMobsCore.java
Hypocisy: Yes, i will change it to non magic string

InfernalMobs/gradle.properties
Hypocisy: Yes it’s specific version of champions, we are using the curse maven, this maven is providing by curse forge, you can find any of the curseforge mod jar to import like champions unofficial: https://www.curseforge.com/minecraft/mc-mods/champions-unofficial/files/6136541
Here to copy the full line: 
IMG_6279
implementation fg.deobf("curse.maven:champions-unofficial-1074990:6133394")

so 1074990 is cf project id, 6133394 is file id, the project id is hard coded since we don’t really change this, the file id is stored by the variable cf_champions

But we just wanted this magic number 6133394

so paste it to
cf_champions=

you can add more depends like this way for better and easy maintenance.
And rebuild or reimport project, then champions(any project) will download automatically.

InfernalMobs/build.gradle
Hypocisy: This line is not needed, I will make a change to fix this.
and yes, if you want other mod that contain mixin to run in dev for test, you need to add those lines to avoid mixin  inject error, that will stop you  build own mod

InfernalMobs/src/main/java/atomicstryker/infernalmobs/common/EntityEventHandler.java
Hypocisy: Alright, let’s add a config to let user choose enable or not enable compatibility, add config let user choose the entity becomes inf or champions, or add chance to let game choose it self.

InfernalMobs/build.gradle
Hypocisy: It’s a soft depend with champions I just used it for test, but it does not need in this pr, will fixed later

InfernalMobs/src/main/java/atomicstryker/infernalmobs/common/InfernalMobsConfig.java
```diff
@@ -185,4 +189,36 @@ public List<String> getEntityTagBlackList() {
     public void setEntityTagBlackList(List<String> entityTagBlackList) {
         this.entityTagBlackList = entityTagBlackList;
     }
+
+    public boolean getUseInfernalMobs() {
+        return useInfernalMobs;
```
AtomicStryker: this name does not describe what the option does very well. "championMobCanBecomeInfernal" would be better, for example

InfernalMobs/src/main/resources/changelog.txt
```diff
@@ -1,3 +1,7 @@
+1.20.1.5
+- Add Champions Compat, if entity already champion mob, stop process spawning entity to inf mob
```
AtomicStryker: best mention each new config value explicitly in the changelog

AtomicStryker (changes requested):
already looking much better

InfernalMobs/src/main/java/atomicstryker/infernalmobs/common/InfernalMobsConfig.java
Hypocisy: I wanted to create a separate config file specifically for mod compatibility settings. However, after completing it, I found that the config file was never generated in the expected location, so I gave up and reverted to using the original file. But I forgot to rename it at that time. Later, when I did rename it, I didn’t understand why it was still not recognized correctly.

InfernalMobs/src/main/resources/changelog.txt
Hypocisy: i forgot, i will write now

InfernalMobs/src/main/java/atomicstryker/infernalmobs/common/InfernalMobsConfig.java
AtomicStryker: once a runtime has generated a config file, you will have to delete it when making code changes. otherwise the gson deserializer will probably stick with defaults and ignore unknowns

EDIT: For the record, i dont use forge/mc provided config files because the stupid things change every 2 versions and i am sick of refactoring. this gson thing does not ever change until i change it

Hypocisy: I tried to delete, but only gen the default one, i use the same way as defaut config, but never work, weird. let's just use the default config lol

InfernalMobs/src/main/resources/changelog.txt
Hypocisy: i writed, please check it

InfernalMobs/src/main/resources/changelog.txt
```diff
@@ -1,3 +1,15 @@
+1.20.1.5
+- Added Champions Compat
+- Added flexible config allow user control mob will handel by champions or Infernal mob, or handel by both
+- Updated the Champions mod to use the new API, allowing it to track whether an entity is a champion and independently manage affixes, rank, and entity modifiers.
+- Added `enableChampionCompat` (boolean) to toggle compatibility with the Champions mod.(default true)
+- Added `useInfernalMobs` (boolean) to allow or prevent Infernal Mobs from handling champion entities.(default true)
```
AtomicStryker: please rename this to "infernalMobsReplacesChampions" as this more correctly describes what it does

InfernalMobs/src/main/resources/changelog.txt
```diff
@@ -1,3 +1,15 @@
+1.20.1.5
+- Added Champions Compat
+- Added flexible config allow user control mob will handel by champions or Infernal mob, or handel by both
```
AtomicStryker: will be handled, handled by both

InfernalMobs/src/main/resources/changelog.txt
Hypocisy: okay

InfernalMobs/src/main/resources/changelog.txt
```diff
@@ -1,3 +1,15 @@
+1.20.1.5
+- Added Champions Compat
+- Added flexible config allow user control mob will be handled, handled by both.
+- Updated the Champions mod to use the new API, allowing it to track whether an entity is a champion and independently manage affixes, rank, and entity modifiers.
+- Added `enableChampionCompat` (boolean) to toggle compatibility with the Champions mod.(default true)
+- Added `infernalMobsReplacesChampions` (boolean) to allow or prevent Infernal Mobs from handling champion entities.(default true)
+- Added `useRandomChampionSelection` (boolean) to enable probability-based entity handling.(default false)
+- Added `championSelectionChance` (double, 0-100) to configure the probability of Champions handling entities instead of Infernal Mobs.(default 50.0)
+- Added `allowBothChampionAndInfernal` to configure mob handel by both mod(default false), for masochists.(default false)
```
AtomicStryker: "mob being changed by both mods", and you write default twice

Hypocisy: sorry, my english is not good

AtomicStryker: it is perfectly sufficient, and i will point out grammar issues i want fixed ;)

AtomicStryker:
i checked out the merged version, built it, tested it locally, looked good to me. great job!

AtomicStryker:
and 1.20.1.5 is pushed to curse. it will take a little while and then show up there

Hypocisy:
> i checked out the merged version, built it, tested it locally, looked good to me. great job!

thanks for your time, you are so kind and good guy

Hypocisy:
> and 1.20.1.5 is pushed to curse. it will take a little while and then show up there

yes, Cheers!!!

PR #556, opened 2025-11-13
Added littletiles combat

CreativeMD:


MultiMine/src/main/java/atomicstryker/multimine/common/MultiMineConfig.java
```diff
@@ -14,6 +14,13 @@ public class MultiMineConfig {
     private boolean disableForAllTileEntities = false;
     private boolean disableAutoRegisterNames = false;
 
+    public MultiMineConfig() {
+        this.bannedBlocks.put("littletiles:tiles", true);
```
AtomicStryker, line 19: please put a comment with a URL to the issue tracker so there is something to read to understand why these are defaults, please

